PR #3, opened 2019-06-03
Testscript update

SimonWan:
1. Update test script;
2. Remove all uint32_t and uint64_t since they are deprecated.


mssun:
Merged, thanks!

PR #24, opened 2021-02-22
Upgrade to OPTEE 3.11.0&3.12.0

DemesneGH:
Hi @mssun ,
As we discussed before, we have upgraded the sdk to OPTEE 3.11.0 and it passed the ci test locally.

The changes we made are:
1. Upgrade optee submodules to 3.11.0(in optee/)
2. Modify some code in optee-utee:
    1)change name of utee syscalls(in tee_api_defines.rs,utee_syscalls.rs,trace.rs).
    2)add -O0 in .cargo/config to avoid generating ta elf with nbuckets==0(which causes ldelf panic).
    3)add definition of __ta_entry(in user_ta_header.rs,tee_api_private.rs,.cargo/config,examples/xxx/ta/ta_static.rs).
    4)change examples/xxx/ta/ta_aarch64.lds to ensure GOT is writable.
3. Modify some code in these examples:
    •authentication
    •diffie_hellman
    •random
    •secure_storage

The new sdk is compatible with OPTEE 3.11.0&3.12.0.Built TAs can run on OPTEE OS 3.8.0,3.11.0 and 3.12.0.Other versions have not been tested.

Contributors are Rong Fan[@fanrong1992](https://github.com/fanrong1992) and Yuan Zhuang[@DemesneGH](https://github.com/DemesneGH).

Best regards,
Yuan Zhuang


.cargo/config
```diff
@@ -2,10 +2,12 @@
 linker = "optee/toolchains/aarch64/bin/aarch64-linux-gnu-ld"
 ar = "optee/toolchains/aarch64/bin/aarch64-linux-gnu-gcc-ar"
 rustflags = [
+	"-C", "link-arg=-e__ta_entry",
```
mssun, line 5: Replace with 4 spaces and align with others.

.cargo/config
```diff
@@ -17,6 +19,7 @@ ar = "optee/toolchains/aarch64/bin/aarch64-linux-gnu-gcc-ar"
 linker = "optee/toolchains/aarch32/bin/arm-linux-gnueabihf-ld.bfd"
 ar = "optee/toolchains/aarch32/bin/arm-linux-gnueabihf-ar"
 rustflags = [
+	"-C", "link-arg=-e__ta_entry",
```
mssun, line 22: Replace with 4 spaces and align with others.

examples/diffie_hellman/host/src/main.rs
```diff
@@ -4,7 +4,13 @@ use proto::{Command, KEY_SIZE, UUID};
 
 fn generate_key(session: &mut Session) -> Result<(Vec<u8>, Vec<u8>)> {
     // Pass in the prime and base
-    let p0 = ParamValue::new(23, 5, ParamType::ValueInput);
+    //let p0 = ParamValue::new(23, 5, ParamType::ValueInput);
```
mssun, line 7: Delete this commented line.

examples/diffie_hellman/ta/src/main.rs
```diff
@@ -44,23 +44,19 @@ fn destroy() {
...
-    let mut key_prime = BigInt::new(64);
-    let mut key_base = BigInt::new(64);
-    key_prime.convert_from_s32(prime_u32 as i32);
-    key_base.convert_from_s32(base_u32 as i32);
-
-    let prime_vec = key_prime.convert_to_octet_string().unwrap();
-    let attr_prime = AttributeMemref::from_ref(AttributeId::DhPrime, prime_vec.as_slice());
-    let base_slice = key_base.convert_to_octet_string().unwrap();
-    let attr_base = AttributeMemref::from_ref(AttributeId::DhBase, base_slice.as_slice());
+	let prime_base_vec = p0.buffer();
```
mssun, line 53: Use 4 spaces instead of tab.

examples/random/host/src/main.rs
```diff
@@ -9,6 +9,7 @@ fn random(session: &mut Session) -> optee_teec::Result<()> {
 
     println!("Invoking TA to generate random UUID...");
     session.invoke_command(Command::RandomGenerator as u32, &mut operation)?;
+	println!("Invokding done!");
```
mssun, line 12: 1. use 4 spaces
2. `Invokding` => `Invoking`

examples/random/ta/src/main.rs
```diff
@@ -31,7 +32,14 @@ fn destroy() {
 
 pub fn random_number_generate(params: &mut Parameters) -> Result<()> {
     let mut p = unsafe { params.0.as_memref().unwrap()};
-    Random::generate(p.buffer());
+	let mut buf = vec![0; p.buffer().len() as usize];
+	buf.clone_from_slice(p.buffer());
+
+	unsafe {
+		raw::TEE_GenerateRandom(buf.as_ptr() as _, p.buffer().len() as _);
```
mssun, line 39: Why using raw API?

examples/secure_storage/ta/src/main.rs
```diff
@@ -127,14 +134,16 @@ pub fn read_raw_object(params: &mut Parameters) -> Result<()> {
                 p1.set_updated_size(obj_info.data_size());
                 return Err(Error::new(ErrorKind::ShortBuffer));
             }
-
-            let read_bytes = object.read(p1.buffer()).unwrap();
+
```
mssun, line 137: Remove unnecessary spaces.

examples/secure_storage/ta/src/main.rs
```diff
@@ -127,14 +134,16 @@ pub fn read_raw_object(params: &mut Parameters) -> Result<()> {
...
-            let read_bytes = object.read(p1.buffer()).unwrap();
+            
+            let read_bytes = object.read(&mut data_buffer).unwrap();
 
             if read_bytes != obj_info.data_size() as u32 {
                 return Err(Error::new(ErrorKind::ExcessData));
             }
 
             p1.set_updated_size(read_bytes as usize);
+            unsafe { std::ptr::copy(data_buffer.as_ptr() as _, p1.buffer().as_ptr() as _, p1.buffer().len() as usize) };
```
mssun, line 145: Why using `std::ptr::copy` instead of `copy_from_slice`?

examples/authentication/ta/src/main.rs
```diff
@@ -105,25 +109,48 @@ pub fn encrypt_final(digest: &mut AEOp, params: &mut Parameters) -> Result<()> {
     let mut p0 = unsafe { params.0.as_memref().unwrap() };
     let mut p1 = unsafe { params.1.as_memref().unwrap() };
     let mut p2 = unsafe { params.2.as_memref().unwrap() };
-    let clear = p0.buffer();
-    let ciph = p1.buffer();
-    let tag = p2.buffer();
-    match digest.op.encrypt_final(clear, ciph, tag) {
+    
+    let mut clear = vec![0; p0.buffer().len() as usize];
+    clear.clone_from_slice(p0.buffer());
```
mssun, line 114: Why not using `copy_from_slice`?

examples/authentication/ta/src/main.rs
```diff
@@ -105,25 +109,48 @@ pub fn encrypt_final(digest: &mut AEOp, params: &mut Parameters) -> Result<()> {
...
+    ciph.clone_from_slice(p1.buffer());
+    let mut tag = vec![0; p2.buffer().len() as usize];
+    tag.clone_from_slice(p2.buffer());
+
+    match digest.op.encrypt_final(&clear, &mut ciph, &mut tag) {
+
         Err(e) => Err(e),
-        Ok((_ciph_len, _tag_len)) => Ok(()),
+        Ok((_ciph_len, _tag_len)) => {
+            unsafe { std::ptr::copy(clear.as_ptr() as _, p0.buffer().as_ptr() as _, p0.buffer().len() as usize) };
```
mssun, line 124: Why not using `copy_from_slice`?

examples/authentication/ta/src/main.rs
```diff
@@ -105,25 +109,48 @@ pub fn encrypt_final(digest: &mut AEOp, params: &mut Parameters) -> Result<()> {
...
+    let mut ciph = vec![0; p1.buffer().len() as usize];
+    ciph.clone_from_slice(p1.buffer());
+    let mut tag = vec![0; p2.buffer().len() as usize];
+    tag.clone_from_slice(p2.buffer());
+
+    match digest.op.decrypt_final(&clear, &mut ciph, &tag) {
         Err(e) => Err(e),
-        Ok(_clear_len) => Ok(()),
+        Ok((_clear_len)) => {
+            unsafe { std::ptr::copy(clear.as_ptr() as _, p0.buffer().as_ptr() as _, p0.buffer().len() as usize) };
```
mssun, line 148: Why not using `copy_from_slice`?

examples/random/ta/src/main.rs
```diff
@@ -31,7 +32,14 @@ fn destroy() {
...
+		std::ptr::copy(buf.as_ptr() as _, p.buffer().as_ptr() as _, p.buffer().len() as usize);
```
mssun, line 40: Why not using `copy_from_slice`?

mssun:
Thanks for your contributions. @xiangmy, can you help to review this PR? Thanks.

optee-utee/optee-utee-sys/src/user_ta_header.rs
```diff
@@ -16,7 +17,18 @@ pub struct ta_head {
...
-    pub entry: unsafe extern "C" fn(c_ulong, c_ulong, *mut utee_params, c_ulong),
+    pub depr_entry: u64,
+}
+
+extern "C" {
+    pub fn __utee_entry(func: c_ulong, session_id: c_ulong, up: *mut utee_params, cmd_id: c_ulong) -> TEE_Result;
+}
+
+#[no_mangle]
+pub fn __ta_entry(func: c_ulong, session_id: c_ulong, up: *mut utee_params, cmd_id: c_ulong) {
```
xiangmy, line 28: `__ta_entry` is a diverging function. Perhaps, this line needs to change to:
```
pub fn __ta_entry(func: c_ulong, session_id: c_ulong, up: *mut utee_params, cmd_id: c_ulong) -> ! {
```

Reference: https://github.com/OP-TEE/optee_os/blob/master/ta/arch/arm/user_ta_header.c#L43

optee-utee/optee-utee-sys/src/utee_syscalls.rs
```diff
@@ -2,10 +2,10 @@ use super::*;
 use libc::*;
 
 extern "C" {
-    pub fn utee_return(ret: c_ulong);
-    pub fn utee_log(buf: *const c_void, len: size_t);
-    pub fn utee_panic(code: c_ulong);
-    pub fn utee_get_property(
+    pub fn _utee_return(ret: c_ulong);
```
xiangmy, line 5: `_utee_return` is also a non-return function. Maybe we should change this to:
```
pub fn _utee_return(ret: c_ulong) -> !;
```

Reference: https://github.com/OP-TEE/optee_os/blob/master/lib/libutee/include/utee_syscalls.h#L32

examples/secure_storage/ta/src/main.rs
```diff
@@ -74,15 +74,19 @@ pub fn delete_object(params: &mut Parameters) -> Result<()> {
 pub fn create_raw_object(params: &mut Parameters) -> Result<()> {
     let mut p0 = unsafe { params.0.as_memref().unwrap() };
     let mut p1 = unsafe { params.1.as_memref().unwrap() };
+
     let mut obj_id = vec![0; p0.buffer().len() as usize];
     obj_id.clone_from_slice(p0.buffer());
+    let mut data_buffer = vec![0; p1.buffer().len() as usize];
+    data_buffer.clone_from_slice(p1.buffer());
```
DemesneGH, line 81: @mssun In original code obj_id uses "clone_from_slice".So is it necessary to use "clone_from_slice", or it could be "copy_from_slice" too?

xiangmy:
Hi @DemesneGH @mssun 

Thanks for the work! The `__ta_entry` and `__utee_return` are diverging functions, so I think we should reflect that. The reset changes look good to me.

BTW, I also add `"-C", "link-arg=max-page-size=4096"` in the `.cargo/config` to force the TA to align at 4 KB. But, I guess the `-O0` flag solves this problem as well as the ldelf panic which I don't know how to deal with. I suppose there would be some better options instead of removing the optimization.

Anyway, thanks for the good work!
Mingyuan

examples/secure_storage/ta/src/main.rs
mssun: Yes.

DemesneGH: OK.I'll update this.

DemesneGH:
Hi @xiangmy 

> But, I guess the `-O0` flag solves this problem as well as the ldelf panic which I don't know how to deal with. I suppose there would be some better options instead of removing the optimization.

The `-O0` flag has the same effect as setting `opt-level=1` in `examples/xxx/ta/Cargo.toml`.Since the default value is `opt-level=3` for rust `release` binaries, it will add a `-O1` when linking the TA elf.Maybe there is a bug of ld which leads invalid `nbuckets` under some circumstances, or OPTEE OS should check the value of `nbuckets`.
Since the [link.mk](https://github.com/OP-TEE/optee_os/blob/3.11.0/ta/arch/arm/link.mk) for TA in OPTEE examples written in C doesn't have the optimization too, we removed the optimization to solve the problem.If there are better solutions we will update then.

Best Regards,
Yuan

xiangmy:
Hi @DemesneGH ,

The `-O0` gives no optimization, and it's corresponding to `opt-level=0`. The OPTEE uses `-Os` by default (see [here](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/arm.mk#L157)). I'm not sure about the default `opt-level` for this sdk. Maybe @mssun can help clarify this.

Thanks,
Mingyuan

optee-utee/optee-utee-sys/src/user_ta_header.rs
DemesneGH: Thanks.I'll update this.

DemesneGH:
Hi @xiangmy 
We printed link args by add `"-Z", "print-link-args"` then found this sdk uses `opt-level=3` for `release` binaries defaultly(same as default profiles [here](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles)), which leads adding `-O1`.When setting `opt-level=1`(or `opt-level=0`) there are no `-Ox` arguments,which is same as `-O0`.
I didn't see `-Os` in printed link args, but I'll try to add `-Os` to find the difference.

Best regards,
Yuan

DemesneGH:
Hi @xiangmy 

> The OPTEE uses `-Os` by default (see [here](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/arm.mk#L157)).

This mk file is probably for OPTEE OS binaries(such as tee.elf), not for TAs.Link.mk for TA is in [here](https://github.com/OP-TEE/optee_os/blob/3.11.0/ta/arch/arm/link.mk),which is a part of `ta_dev_kit`(imported when building TAs),and it seems not having `-Os`.

Best Regards
Yuan

DemesneGH:
Hi @mssun 
Thanks for your review.I've updated the code.

Best Regards,
Yuan

xiangmy:
> Hi @xiangmy
> 
> > The OPTEE uses `-Os` by default (see [here](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/arm.mk#L157)).
> 
> This mk file is probably for OPTEE OS binaries(such as tee.elf), not for TAs.Link.mk for TA is in [here](https://github.com/OP-TEE/optee_os/blob/3.11.0/ta/arch/arm/link.mk),which is a part of `ta_dev_kit`(imported when building TAs),and it seems not having `-Os`.
> 
> Best Regards
> Yuan

Hi @DemesneGH ,

TAs have the same optimization level as the OS. The flow of passing the `-Os` flag is (for aarch64):
[`platform-cflags-optimization`](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/arm.mk#L159) -> [`ta_arm64-platform-cflags`](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/arm.mk#L276) -> [`ta-mk-file-export-vars-ta_arm64`](https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/arm.mk#L295) -> [`ta.mk`](https://github.com/OP-TEE/optee_os/blob/master/ta/ta.mk)

Would adding `-Os` (instead of `-O0`) and `opt-level = "s" ` work in our case?

Thanks,
Mingyuan

DemesneGH:
Hi @xiangmy 

> Would adding `-Os` (instead of `-O0`) and `opt-level = "s" ` work in our case?

Yes.Adding `-Os` instead of `-O0` works, and set `opt-level="s"`will also add `-Os` in link arguments.
I‘ve tried several cases and the results are:

1. Setting `opt-level="s"` will add `-Os`,builds elf of `nbuckets` == 1
2. Setting `opt-level=3` will add `-O1`,builds elf of `nbuckets` == 0
3. Setting `opt-level=2` will add `-O1`,builds elf of `nbuckets` == 0
4. Setting `opt-level=1` will not add `-Ox`(`-O0` by default),builds elf of `nbuckets` == 1
5. Setting `opt-level=0` will not add `-Ox`(`-O0` by default),builds elf of `nbuckets` == 1


This [man page](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html) says :

> -Os enables all -O2 optimizations except those that often increase code size:
-falign-functions  -falign-jumps 
-falign-labels  -falign-loops 
-fprefetch-loop-arrays  -freorder-blocks-algorithm=stc

But it's still unclear what the optimizations do and what leads illegal `nbuckets`.
Anyway,thanks for your advice.I'll update `-O0` to `-Os`.

Best regards,
Yuan

mssun:
Merged. Thank you all!

PR #29, opened 2021-06-21
Add support for tee-supplicant loadable plugins

DemesneGH:
Hi @mssun 
The loadable plugins framework is a new feature in OPTEE 3.13.0. It allows TA actively invoke a command defined in normal world plugins (which are compiled as .so dynlibs).See [this commit](https://github.com/OP-TEE/optee_client/commit/1e91cc797f72261fb03e7c2ce30944ff613ac07e).
In this commit I've updated:
1. In`optee-teec` `optee-utee` and `optee-utee-sys`:Wrap for related apis;
2. Add a new example `supp_plugin` for tee-supplicant plugins.The work logic is `host->TA->plugin`.Plugins are identified by uuid.

Best Regards,
Yuan


examples/supp_plugin/plugin/plugin_static.rs
```diff
@@ -0,0 +1,27 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#[no_mangle]
+pub static mut plugin_method: PluginMethod = PluginMethod {
```
m4sterchain, line 19: What is the intention to make this data structure mutable? It seems that the [tee_supplicant logic](https://github.com/OP-TEE/optee_client/blob/7c9c423d00e96bf51debd5fe10fd70dce83be5cc/tee-supplicant/src/plugin.c#L85) only needs a constant symbol to look up.

DemesneGH: After removing the `mut`, this error occurs: 
```
error[E0277]: `*mut u8` cannot be shared between threads safely.
  = help: within `optee_teec::extension::PluginMethod`, the trait `std::marker::Sync` is not implemented for `*mut u8`
   = note: required because it appears within the type `optee_teec::extension::PluginMethod`
   = note: shared static variables must have a type that implements `Sync`
```

mssun:
Merged. Thanks!

PR #36, opened 2021-09-01
Update makefile and environment variables

DemesneGH:
- .cargo/config: remove relative path of toolchains since it has been added to `PATH` in `source environment`
- Makefile: add installation of plugins in example-install
- Makefile, environment and README.md: `OPTEE_DIR` allows setting OPTEE root path
- .github/workflows/ci.yml: fix related ci error


mssun:
Thanks!

PR #38, opened 2021-09-07
Fix ldelf error of arm32 ta

DemesneGH:
- `examples/*/ta/ta_arm.lds`: Add write permisson of GOT
- `examples/*/ta/Cargo.toml`: Set `opt-level=1` to compile arm32 ta of `nbuckets!=0`. In #24 we had updated `linkarg=-Os` and  `opt-level=3`(by default) which worked for arm64 ta but didn't work for arm32 ta.There should be  `linkarg=-Os` and  `opt-level=1` now.


mssun:
Merged. Thanks!

PR #73, opened 2022-05-06
Add signature verification example

DemesneGH:
- Add the example of signing a message and verifying the signature using the third party crate [ring](https://github.com/veracruz-project/ring.git).
- Add the test for signature_verification-rs in `tests/` and `qemu-check.exp`.


mssun (approved):
LGTM, thanks!

PR #77, opened 2022-05-12
Add tls server example

DemesneGH:
The `tls_server-rs` example sets up the TLS server in TA. Same as the `tls_client-rs` example, it also needs the larger TA heap buffer. So in the CI process, it uses QEMU images which expand the TA memory.

Updates:
- Add examples/tls_server-rs/
- Add tests/test_tls_server.sh
- Update tests/optee-qemuv8.sh for forwarding the host port `54433` to the guest port `4433`
- Ignore keys and certs in the license checking process


DemesneGH:
@mssun There are more keys and certs in `tls_server-rs/ta/test-ca` which are generated in multiple algorithms (`ecdsa`, `eddsa`, and `rsa`). But the tls server example only loads `ecdsa` keys.  Should we remove the `build-a-pki.sh` script and the other keys? Or just reserve them in the directory?

mssun:
> @mssun There are more keys and certs in `tls_server-rs/ta/test-ca` which are generated in multiple algorithms (`ecdsa`, `eddsa`, and `rsa`). But the tls server example only loads `ecdsa` keys. Should we remove the `build-a-pki.sh` script and the other keys? Or just reserve them in the directory?

I think we can remove the script and leave the test certificates only.

mssun (approved):
LGTM, thanks!

DemesneGH:
> I think we can remove the script and leave the test certificates only.

OK. The script `build-a-pki.sh` and the related config file `openssl.cnf` have been removed.

PR #104, opened 2023-03-07
Update to OP-TEE 3.20.0

DemesneGH:
After OP-TEE 3.20.0, OP-TEE updated the GP support from 1.1 to 1.3.1 (https://github.com/OP-TEE/optee_os/pull/5688), which made some of the libutee APIs not compatible with the current Rust wrappers.
So in `ci.yml` I've pinned the OP-TEE repo to `3.20.0`.

Notes: todo list before updating to OP-TEE 3.21.0:
- re-generate Rust wrappers for libutee in `optee-utee/optee-utee-sys/src/`: `tee_api.rs`, `utee_syscalls.rs`, etc.
- update the corresponding function calls in `optee-utee/src`.
- ci: resume the OP-TEE repo to the latest version.


henrysun007:
Two many test files are modified. Putting the version number in one file as a macro and referencing it in other files may reduce the modified files when there is a new version number.

DemesneGH:
@henrysun007 Thanks for your comment. Since this PR was merged, I'll update those test scripts before the next version (OP-TEE 3.21.0).

PR #105, opened 2023-03-13
Update to GP 1.3.1

DemesneGH:
According to the GP 1.3.1 updates in OP-TEE: https://github.com/OP-TEE/optee_os/pull/5688. There are the items that  may be updated in Trustzone SDK correspondingly: 

items we have finished:

- add some new constant defines in `tee_api_defines.rs`
  
- update all size-related types in `TEE_*` APIs from `u32` to `usize`
  
- update the parameter name of `TEE_AllocateTransientObject` from `maxKeySize` to `maxObjectSize`
  
- fix `supp_plugin-rs` panic. It panics because the new check has been added in OP-TEE: buffers read-only memrefs should not be passed on to other TAs as read/write
  

items that we've not added/implemented, but not affect our current functionalities:

- `TEE_ALG_*` , `TEE_ATTR_X448_*` , `TEE_ATTR_HKDF_*` defines
  
- two new values of the `hint` paramter in `TEE_Malloc()`and `TEE_Realloc()`: `TEE_MALLOC_NO_FILL` and `TEE_MALLOC_NO_SHARE`
  
- the new TA properties `gpd.tee.internalCore.version`, `gpd.ta.endian`, `gpd.client.endian`, and `gpd.ta.doesNotCloseHandleOnCorruptObject`
  
- APIs for the `SHA-3` algorithms: `SHA3-224`, `SHA3_256`, `SHA3_384`,
  `SHA3_512`, `SHAKE128`, and `SHAKE256`
  
- Implement big int APIs: `TEE_BigIntAssign()`, `TEE_BigIntSetBit()`, `TEE_BigIntAbs()`, and `TEE_BigIntExpMod()`

Other things that should be noted:
- Because our CI runs on the pre-built OP-TEE binaries of release versions, so I've temporarily disabled it until the OP-TEE 3.21.0 is released. We are using the CI in the latest OP-TEE repo for now.
-  The GCC toolchains have been updated from `10.2` to `11.3.rel1` after OP-TEE 3.20.0. The TAs built on the new version of toolchains cannot run on the old QEMU rootfs because of the inconsistency of GCC libraries.


mssun (approved):
LGTM. Thanks!

PR #107, opened 2023-04-13
Add "links" field to Cargo.toml for teec and utee

tchebb:
This lets dependents with their own way of linking to libteec and/or libutee override our build script if they so choose. See the documentation [here](https://doc.rust-lang.org/cargo/reference/build-scripts.html#overriding-build-scripts).

This contribution is on behalf of my company.


DemesneGH:
LGTM. Thanks!

PR #108, opened 2023-04-21
optee-teec: update OP-TEE implementation fields

tchebb:
We've been updating our OP-TEE version without updating our Rust definitions of its API types. Since the GlobalPlatform TEE Client API uses client-allocated API objects, our code needs to be aware of and match the implementation-specific fields used by the version of libteec we link to.

This pulls in two changes that we missed, one of which increases the space we need to allocate for TEEC_Context and so results in memory unsafety with our current definitions:

 - OP-TEE/optee_client#145
 - OP-TEE/optee_client#217

### DIFF
--- a/optee-teec/optee-teec-sys/src/tee_client_api.rs
+++ b/optee-teec/optee-teec-sys/src/tee_client_api.rs
@@ -80,6 +80,7 @@ pub type TEEC_Result = u32;
 pub struct TEEC_Context {
     pub fd: c_int,
     pub reg_mem: bool,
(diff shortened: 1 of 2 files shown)


tchebb:
Sorry, recreated the branch so the CI would run. Didn't realize that would auto-close the PR.

mssun:
@tchebb Thanks for the contribution. Sometimes, the CI pipeline needs approval to run.

DemesneGH (approved):
LGTM. Thanks!

PR #109, opened 2023-04-21
`setup.sh` and default toolchain fixes

tchebb:
Several fixes aimed at making dev environment setup more reliable and friendly for users new to the project. See individual commit messages for details. Don't squash when merging if possible.


.github/workflows/ci.yml
```diff
@@ -92,6 +86,11 @@ jobs:
           mkdir -p ~/optee-qemuv8 && cd ~/optee-qemuv8 &&
           repo init -u https://github.com/OP-TEE/manifest.git -m qemu_v8.xml &&
           repo sync -j4 --no-clone-bundle
+      - name: Setting up $HOME
+        run: |
+          cp /root/.bashrc $HOME/.bashrc &&
+          ln -sf /root/.rustup ~/.rustup &&
+          ln -sf /root/.cargo ~/.cargo
```
DemesneGH, line 93: Seems that CI works fine without setting that environment in previous running jobs, so it's really needed to do this?

tchebb: I think it's because we unconditionally installed our own version of rustup from https://rustup.rs/ before. My changes to `setup.py` make us use an installed version if it exists, though, which means that installed version has to work. And the version in the GitHub Actions container we use doesn't seem to work unless those symlinks exist.

tchebb:
(Latest push didn't change any code, just the commit message of "CI: use home directory workaround for tests too" which had previously erroneously referred to a self-hosted runner.)

DemesneGH (approved):
LGTM, thanks!

PR #120, opened 2024-02-03
Support `Send`able sessions for use in multithreaded code

maurer:


DemesneGH:
The CI has been fixed, please rebase on the latest commit, thanks!

optee-teec/src/session.rs
```diff
@@ -40,16 +44,58 @@ pub enum ConnectionMethods {
     LoginGroupApplication,
 }
 
+// Sessions are documented as threadsafe in GP. Sessions with non-static lifetimes
+// cannot be send because we cannot ensure that their context lives long enough, but
+// sessions with 'static lifetime should be Send.
+//
+// A few ways to construct a session with static lifetime:
+// 1. With the "owned" feature, use an `OwnedSession`
```
DemesneGH, line 52: Can we have a `hello-world` example for multi-thread session? It would be great if the example can be tested on our CI

b49020:
@maurer Can you target these changes for `no-std` branch instead since that will be the main development branch going forward?

@DemesneGH I hope that's something we agreed upon. We will merge `std` TAs support there once the custom downstream OP-TEE toolchain target makes it way upstream.

DemesneGH:
@b49020 For previous users developed on the `master`(with `std`) it's okay to open PR upon this branch. If it's ready to be merged we can sync the change to `no-std` branch then. I think developers should not be tangled in open PRs on which branch, we can handle that.

b49020:
@DemesneGH I think the better way to handle this is to rather create separate git repos for `optee-utee` and `optee-teec` crates which both `std` and `no-std` examples can reuse. All the contributions happens towards those git repos instead, see: https://github.com/apache/incubator-teaclave-trustzone-sdk/pull/123. It would also allow us to easily publish those crates to crates.io too.

x1a0b0:
Any update to this PR?
It's meaningful for integrate CA to normal rust applictions with multithread support

DemesneGH:
@x1a0b0 Thanks for your contribution! Since the `master` branch is outdated, could you rebase on the `main` branch and open a new PR? Please also see the review comments I left before, thanks!

ivila:
Actually we have done the same thing too, our use case is to build a session pool by using `mobc` or `r2d2`, and both of them require the Session struct type to be `Send + 'static`, currently we use [ouroboros](https://github.com/someguynamedjosh/ouroboros.git) to build a self reference struct (like what @maurer do with self_cell):
```rust
#[ouroboros::self_referencing]
pub struct Session {
    ctx: optee_teec::Context,
    #[borrows(mut ctx)]
    #[covariant]
    sess: optee_teec::Session<'this>,
}
```

On my opinion, if we want to introduce a `'static Session` officially, we should just stop using crate like ouroboros and self_cell to make a self reference patch, but just introduce a Session struct holding both `optee_utee_sys::Context` and `optee_utee_sys::Session`, rather than using a share Context wrapper with Arc, because it can easily leads to bugs in rust, and currently there is a bug in this PR, it change the declaration of parameter context in Session::new from `&'ctx mut Context` to `&'ctx Context`, which can lead to race condition.
Currently implementation of teaclave:
image
And in this PR, it changes to:
image
if you move `context.as_mut_raw_ptr()` out of unsafe, rust compiler will give it a compile error, because parameter context is not a mutable borrow.

Actually, currently teaclave use a mutable borrow when calling TEEC_OpenSession is a correct way to go, because OP-TEE doesn't promise Context to be concurrent safe when using as parameter of TEEC_OpenSession, so it it required to get a mutable borrow to prevent any other concurrent visit.

> PS: in rust, mutable doesn't just meaning a value can be changed, but also preventing concurrency visit, and immutable doesn't mean a value cannot be changed too, if you promise a value must not be visit concurrently(for example, a immutable RWMutex can get its inner value as mutable after lock), you can change it too, that's the point of the memory safe in rust.

PR #123, opened 2024-02-08
RFC: An effort to allow reuse for common OP-TEE crates

b49020:
I would propose to have separate git repos for `optee-utee` and `optee-teec` crates such that they can be reused by both `std` and `no-std` OP-TEE Rust examples. This would allow us to avoid divergent OP-TEE crates and rather have single git repos to provide collaborative development environment.

For demonstration purpose, I have created two repos using `git subtree split` command as follows:
- https://github.com/b49020/optee-teec
- https://github.com/b49020/optee-utee

@DemesneGH If you are fine with this approach then please create corresponding 2 new repos for these crates and I will create PRs for those. Once those are in place then we can drop `optee-utee` and `optee-teec` sub-directories from both `master` and `no-std` branch.

### DIFF
--- a/environment
+++ b/environment
@@ -24,6 +24,7 @@ fi
 export OPTEE_OS_DIR="$OPTEE_DIR/optee_os"
 export OPTEE_CLIENT_DIR="$OPTEE_DIR/optee_client/out"
 export OPTEE_CLIENT_INCLUDE="$OPTEE_DIR/optee_client/out/export/usr/include"
+export OPTEE_CLIENT_EXPORT="$OPTEE_DIR/optee_client/out/export"
 if [ "$ARCH" = "arm" ]
 then
   export ARCH="arm"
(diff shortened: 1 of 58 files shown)


DemesneGH:
@b49020 Thanks for your proposal!
Let me clarify our goal first:
We absolutely understand the stability and efficiency of `no-std` for OPTEE Rust TAs, and are willing to switch to `no-std` and set it as `default` branch. The `std` feature are also planned to be added upon the that branch.

For the transition period:
Considering some of our legacy developers may don't want to switch to new branch due to maintainability of their projects, we still keep the legacy `master` branch and accept PRs upon that. It would be better if developers contribute on the new default branch but it depends on them . If the update also makes sense on `no-std` branch, we can sync that. 

For speeding up the transition period, can we have some description on our README main page which recommend using no-std and it's benefits? If the `no-std` is ready (seems there are some CI error), feel free to open the PR of updated README, the `default` branch will be set after the documentation is ready. I believe more developers will choose the `no-std` then.

Regarding this PR: 
I understand the benefit of separate repo which would be clean when we publish it on `crate.io` but it's not mandatory for published crates.
I choose not to merge this PR for now because:
1. The "transition period" still exists and the separate repo may not speed that up.
2. The `optee-utee` and `optee-teec` are core modules of `incubator-teaclave-trustzone-sdk` repo, if they're moved outside, the `incubator-teaclave-trustzone-sdk` repo may be useless (examples and tests can also go with the separate crates).
3. Since we are project under the Apache Community, creating a new repo needs to involve them in.

Please correct me if I misunderstand, thanks!

b49020:
With `main` branch supporting `std` and `no-std`, the purpose of this PR is achieved, hence closing.

Issue #135, opened 2024-05-22
Build issues on aarch64 Linux hosts


msgilligan:
I've been trying to build OP-TEE with Rust on an aarch64 Linux host.  (Mostly Debian Bookworm in a VM on MacBook M1)

I've run into at least three issues so far and made a PR for one of them:

* https://github.com/OP-TEE/build/issues/749, with a fix PR https://github.com/OP-TEE/build/pull/748
* https://github.com/OP-TEE/build/issues/751
* https://github.com/OP-TEE/build/issues/752

751 is not Rust-related, and is easily worked around with `WITH_CXX_TESTS=n`, but 752 is Rust-related and if anyone has any ideas on how to fix it, I would greatly appreciate hearing them.

msgilligan:
I have reproduced this issue by following the instructions in this projects `README.md` on an aarch64 Linux (Debian) machine.

msgilligan:
I have created a multi-architecture Dockerfile/containerFile that works correctly under AMD64 (x86_64) and exhibits the linker error we have been seeing under ARM64 (aarch64) and uploaded it in a gist:
https://gist.github.com/msgilligan/a36d61953d2caf60d33d233fed281c7b

msgilligan:
Note that since https://github.com/OP-TEE/build/pull/764 (and related changes) were merged to `OP-TEE/build`  RUST_ENABLE defaults to `n`.

I'm assuming that if we can fix this issue that `OP-TEE/build` can default to `RUST_ENABLE=y` again.

b49020:
Thanks @msgilligan for continuously following this issue. It has been on my ToDo list. BTW, thanks for that docker fine which allowed me to reproduce it.

msgilligan:
> Thanks @msgilligan for continuously following this issue. It has been on my ToDo list. BTW, thanks for that docker fine which allowed me to reproduce it.

I really need this fixed as I'm trying to write a TA in Rust and my main development system is an ARM64 laptop (MacBook Pro M1 with Debian in a VM) so I am motivated to get it fixed. Unfortunately, I don't know much about Rust linking or the custom link options used for TAs.

I'm hoping this get fixed by someone else, but I am willing to help with testing, documentation, or even coding a fix. Is there any documentation about the custom link options for TAs?

ivila:
[...]

PR #137, opened 2024-05-29
Add aarch64 support to environment scripts

msgilligan:
~~This is a DRAFT PR for preliminary feedback.~~  I am trying to resolve Issue #135 and I had to make changes to these two files to get far enough to reproduce the issue.

PR #138, opened 2024-05-29
hack to 'fix' Issue #135 for acipher-rs TA

msgilligan:
This is a "hack" which demonstrates changes that I made that eliminate the error seen in Issue #135. I don't understand the compile/link process well enough to submit a real fix and I haven't even tested that the build of this TA works, but I am submitting this patch as a demonstration of how to make the error message seen in Issue #135 go away.

### DIFF
--- a/examples/acipher-rs/ta/Makefile
+++ b/examples/acipher-rs/ta/Makefile
@@ -20,7 +20,7 @@ UUID ?= $(shell cat "../uuid.txt")
 TARGET ?= aarch64-unknown-linux-gnu
 CROSS_COMPILE ?= aarch64-linux-gnu-
 OBJCOPY := $(CROSS_COMPILE)objcopy
-LINKER_CFG := target.$(TARGET).linker=\"$(CROSS_COMPILE)ld.bfd\"
+LINKER_CFG := target.$(TARGET).linker=\"$(CROSS_COMPILE)ld\"
 
 TA_SIGN_KEY ?= $(TA_DEV_KIT_DIR)/keys/default_ta.pem
 SIGN := $(TA_DEV_KIT_DIR)/scripts/sign_encrypt.py
@@ -29,6 +29,7 @@ OUT_DIR := $(CURDIR)/target/$(TARGET)/release
 all: ta strip sign
 
 ta:
+	@echo ${LINKER_CFG}
 	@cargo build --target $(TARGET) --release --config $(LINKER_CFG)
 
 strip: ta
--- a/examples/acipher-rs/ta/build.rs
+++ b/examples/acipher-rs/ta/build.rs
@@ -94,10 +94,11 @@ fn main() -> std::io::Result<()> {
     println!("cargo:rustc-link-arg=-e__ta_entry");
     println!("cargo:rustc-link-arg=-pie");
     println!("cargo:rustc-link-arg=-Os");
-    println!("cargo:rustc-link-arg=--sort-section=alignment");
+    //println!("cargo:rustc-link-arg=--sort-section=alignment");
+
 
     let mut dyn_list = File::create(out.join("dyn_list"))?;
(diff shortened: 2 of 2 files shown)


b49020:
@msgilligan Since it's a buildroot toolchain compiled from source in case of native aarch64 builds. I would like to see the linker path provided. Can you try following diff instead?

```
diff --git a/examples/acipher-rs/ta/Makefile b/examples/acipher-rs/ta/Makefile
index ad35909..498b22e 100644
--- a/examples/acipher-rs/ta/Makefile
+++ b/examples/acipher-rs/ta/Makefile
@@ -20,7 +20,7 @@ UUID ?= $(shell cat "../uuid.txt")
 TARGET ?= aarch64-unknown-linux-gnu
 CROSS_COMPILE ?= aarch64-linux-gnu-
 OBJCOPY := $(CROSS_COMPILE)objcopy
-LINKER_CFG := target.$(TARGET).linker=\"$(CROSS_COMPILE)ld.bfd\"
+LINKER_CFG := target.$(TARGET).linker=\"$(CROSS_COMPILE)ld\"
 
 TA_SIGN_KEY ?= $(TA_DEV_KIT_DIR)/keys/default_ta.pem
@@ -29,6 +29,7 @@ OUT_DIR := $(CURDIR)/target/$(TARGET)/release
 all: ta strip sign
 
 ta:
+       @echo ${LINKER_CFG}
        @cargo build --target $(TARGET) --release --config $(LINKER_CFG)
 
 strip:
```

msgilligan:
> Since it's a buildroot toolchain compiled from source in case of native aarch64 builds.

After seeing the following line in `setup.sh` in this repository, I have been experimenting with using the standard apt toolchain on an aarch64 host:

```
apt update && apt -y install gcc-aarch64-linux-gnu gcc-arm-linux-gnueabihf
```

When I tried to apt install `gcc-aarch64-linux-gnu` I got the following message:

```
Note, selecting 'gcc' instead of 'gcc-aarch64-linux-gnu'
gcc is already the newest version (4:12.2.0-3).
```

So this lead me to create PR #137.
 
>  Can you try following diff instead?

It's force-pushed and I will try it out in my aarch64 Debian VM.

Please note the two commented-out lines in `examples/acipher-rs/ta/build.rs`. They set the following two linker options, which I suspect are necessary for proper TA operation (but have not tested yet):

```
--sort-section=alignment
--dynamic-list=dyn_list
```

This PR is meant to document my work-in-progress and to stimulate discussion, it is a "very drafty" DRAFT.

Note that I am new to OP-TEE, relatively new to Rust, have not used Rust in embedded environments before and am not knowledgeable about these linker options, so I'm experimenting and learning as I go along...

msgilligan:
> > Can you try following diff instead?
> 
> It's force-pushed and I will try it out in my aarch64 Debian VM.

If I use this PR after my force-push of your diff, I get the following error message:

```
ld: cannot find -lgcc_s: No such file or directory
```

This is the same as with `ld.bfd`.

If I remove the `--config $(LINKER_CFG)` from the end of the `@cargo` line, the build successfully completes. So it seems the `--config $(LINKER_CFG)` must be doing more than just switching the linker command.

Is there a way to find out what else it is doing?

b49020:
Can you share the output of following echo command?

```
+       @echo ${LINKER_CFG}
```

b49020:
> If I remove the `--config $(LINKER_CFG)` from the end of the `@cargo` line, the build successfully completes. So it seems the `--config $(LINKER_CFG)` must be doing more than just switching the linker command.
> 
> Is there a way to find out what else it is doing?

It is there to setup linker for the aarch64 target when Rust examples are being cross-compiled. However, here you are trying a native aarch64 Rust examples build which isn't tested before. Let me see if I can test it on a aarch64 machine too.

msgilligan:
> Can you share the output of following echo command?
> 
> ```
> +       @echo ${LINKER_CFG}
> ```

```
target.aarch64-unknown-linux-gnu.linker="ld"
``

ivila:
Close as already fix by #156

PR #141, opened 2024-07-08
[main] add std support

DemesneGH:
The `main` branch forks from `no-std` and the optional `std` support has been added upon it. After the `main` branch is ready we can set it as the default branch.

The main changes are:

#### To add `std`:
- add `setup_std_dependencies.sh` for cloning Rust src
- add target spec for `aarch64-unknown-optee` and `arm-unknown-optee`
- `optee-utee-sys` and `optee-utee`: remove the default feature `std`
- examples/*/ta/Makefile: using Xargo for building when `STD` env is set

#### Building system adjustment:
- the `no-std` is default and the std mode can be enabled by `$ export STD=y`
- setup env using `$ source environment`: users should set `ARCH_HOST` (default aarch64), `ARCH_TA` (default aarch64) and `STD` (default unset) before building
- `build_optee_libraries.sh` builds libraries for both aarch64 and arm32
- test scripts: read the current `ARCH_HOST` and `ARCH_TA` for fetching binaries
- corresponding ci adjustment

#### Other updates:
- update Rust toolchain to `nightly-2024-05-15`
- examples/hotp/ta: fix panic


DemesneGH:
@b49020 Could you help to review this? thanks!

optee-utee/optee-utee-sys/src/tee_api.rs
```diff
@@ -17,9 +17,9 @@
 
 use super::*;
 
-#[cfg(feature = "std")]
-use std::os::raw::*;
-#[cfg(not(feature = "std"))]
+//#[cfg(target_os = "optee")]
+//use libc::*;
+//#[cfg(not(target_os = "optee"))]
```
b49020, line 22: These can be dropped as I only required then initially since `std` support required old rust toolchain which didn't provide `core::ffi::*;` which isn't the case any longer. Same for all other instances below.

.github/workflows/ci.yml
```diff
@@ -67,39 +67,59 @@ jobs:
...
-          source environment
 
-          # Build OP-TEE Rust examples for Arm 64-bit both host and TA
+          # Build OP-TEE Rust examples for Arm 32-bit both host and TA
+          export ARCH_HOST=arm
+          export ARCH_TA=arm
+          source environment
           make -j`nproc`
 
-          # Build OP-TEE Rust examples for Arm 64-bit host and 32-bit TA
```
b49020, line 78: I see you have dropped mixed build tests with host for 64-bit and TA for 32-bit or vice-versa. Any particular reason to do so?

.github/workflows/ci.yml
```diff
@@ -67,39 +67,59 @@ jobs:
...
+    container: teaclave/teaclave-trustzone-sdk-build:0.3.0
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Setting up $HOME
+        run: |
+          cp /root/.bashrc $HOME/.bashrc &&
+          ln -sf /root/.rustup ~/.rustup &&
+          ln -sf /root/.cargo ~/.cargo
+      - name: Building Arm 64-bit both host and TA (with STD enabled)
```
b49020, line 99: Can we support 32-bit TAs with STD enabled? I can see that we support 32-bit OP-TEE custom target for STD.

examples/acipher-rs/host/Cargo.lock
b49020, line 1: These cargo.lock files can be dropped, riight?

examples/aes-rs/Makefile
```diff
@@ -26,10 +26,10 @@ TARGET_HOST ?= $(TARGET)
 TARGET_TA ?= $(TARGET)
```
b49020, line 26: This default assignment can be dropped with the renaming in this patch-set?

examples/acipher-rs/ta/Makefile
```diff
@@ -17,19 +17,23 @@
...
 SIGN := $(TA_DEV_KIT_DIR)/scripts/sign_encrypt.py
-OUT_DIR := $(CURDIR)/target/$(TARGET)/release
+OUT_DIR := $(CURDIR)/target/$(TARGET_TA)/release
 
 all: ta strip sign
 
 ta:
-	@cargo build --target $(TARGET) --release --config $(LINKER_CFG)
+	@if [ $(STD) ]; then\
+		xargo build --target $(TARGET_TA) --release --config $(LINKER_CFG) --verbose;\
```
b49020, line 33: Do we really need a verbose build in STD case?

optee-utee/Cargo.toml
```diff
@@ -24,18 +24,13 @@ repository = "https://github.com/apache/incubator-teaclave-trustzone-sdk.git"
...
 optee-utee-macros = { path = "macros" }
 bitflags = "=1.0.4"
 uuid = { version = "0.8", default-features = false }
 hex = { version = "0.4", default-features = false, features = ["alloc"] }
 libc_alloc = "=1.0.5"
 
-[workspace]
-members = ['systest']
+#[workspace]
+#members = ['systest']
```
b49020, line 36: Should these be dropped?

rust-toolchain.toml
```diff
@@ -18,5 +18,8 @@
 # Toolchain override for rustup
 
 [toolchain]
-channel = "nightly-2023-12-18"
+channel = "nightly-2024-05-15"
+components = [ "rust-src" ]
```
b49020, line 22: I suppose this won't add too much size increase for `no-std` CI, right?

b49020 (commented):
@DemesneGH This is a welcome change since having two different branches for `std` and `no-std` case was cumbersome to maintain as well as for contributors to contribute changes. BTW, I have a few comments inline.

rust-toolchain.toml
DemesneGH: Building `std` needs the components `rust-src` and it increase the download size indeed, but setting the profile as minimal removes the components we do not need (at line 24-25) which reduces the download size:
```
# minimal profile: install rustc, cargo, and rust-std
profile = "minimal"
```

By default it will download `cargo`, `clippy`, `rust-docs`, `rustc`, `rustfmt`...

.github/workflows/ci.yml
DemesneGH: The host and TA are built separately for each example, seems building both aarch64 and both arm32 cover all cases. If `building host as aarch64 and TA as arm32` doesn't introduce new case I think it's okay to remove them

.github/workflows/ci.yml
DemesneGH: Yes we support 32-bit TAs with std, I'll add a CI job for that.

examples/acipher-rs/host/Cargo.lock
DemesneGH: Rust projects usually keep the `Cargo.lock` for binaries (refernce: https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control) 
But considering we have two building mode, should we keep the `Cargo.lock` for `no-std` building, or just discard all Cargo.lock (maybe lack version control)?
I'd prefer the first choice, any suggestions?

.github/workflows/ci.yml
b49020: The reason for me to add those tests was to be confident that our top level scripts or Makefile does proper required configuration to handle every case. I suppose it won't be much cumbersome for the CI.

examples/acipher-rs/host/Cargo.lock
b49020: There is another perspective to look at it to rather allow verification of latest dependencies in our CI and not checking in `Cargo.lock` files is a potential solution: https://doc.rust-lang.org/cargo/guide/continuous-integration.html#verifying-latest-dependencies.

Another thing I see negative for `Cargo.lock` files is how often do we update them and if we are really leveraging the determinism offered. For an upstream project like this we would rather like to track breakages with any version of upstream Rust packages such that we can update `Cargo.toml` to not depend on versions that are broken.

So IMHO given we already have two configurations to support and the reasons above, we should be better off to drop `Cargo.lock`. However, I think these `Cargo.lock` makes more sense for downstream projects where tight version control is preferred more.

rust-toolchain.toml
b49020: Okay that's fine as long as the CI is happy.

examples/acipher-rs/ta/Makefile
DemesneGH: The verbose is not necessary so it can be simplified as `BUILDER = $(if $(STD),xargo,cargo)`, will update later

examples/aes-rs/Makefile
DemesneGH: Considering `examples/*/host/Makefile` and `examples/*/ta/Makefile` already have default assignment, I think this Makefile can be simplified to "
all:
	@make -s -C host
	@make -s -C ta

clean:
	@make -s -C host clean
	@make -s -C ta clean
", any comments?

b49020: I am sceptical if that would allow us to override the defaults simply via `make TARGET_TA=<>` assignments but rather require explicit `export TARGET_TA=<>`. Please double check.

examples/acipher-rs/host/Cargo.lock
DemesneGH: `Cargo.lock` removed

examples/aes-rs/Makefile
DemesneGH: Considering `make TARGET_TA=<>` it's better to keep those variables. Have updated to 
```
CROSS_COMPILE_HOST ?= aarch64-linux-gnu-
CROSS_COMPILE_TA ?= aarch64-linux-gnu-
TARGET_HOST ?= aarch64-unknown-linux-gnu
TARGET_TA ?= aarch64-unknown-linux-gnu
```

optee-utee/Cargo.toml
DemesneGH: commented out for test, enable now

optee-utee/optee-utee-sys/src/tee_api.rs
DemesneGH: done

.github/workflows/ci.yml
DemesneGH: added

DemesneGH:
Updated for all comments. Could you please review again? Thanks! @b49020

.github/workflows/ci.yml
```diff
@@ -65,41 +65,110 @@ jobs:
...
+          export ARCH_HOST=arm
+          export ARCH_TA=arm
+          source environment
+          make -j`nproc`
+
+          # Build OP-TEE Rust examples for Arm 32-bit host and 64-bit TA
+          export ARCH_HOST=arm
+          unset ARCH_TA
+          source environment
+          make -j`nproc`
```
b49020, line 82: I think this deserves a clean like:

```
make clean && make -j`nproc`
```

.github/workflows/ci.yml
```diff
@@ -65,41 +65,110 @@ jobs:
...
+
+          # Build OP-TEE Rust examples for Arm 64-bit host and 32-bit TA
+          unset ARCH_HOST
+          export ARCH_TA=arm
+          source environment
+          make -j`nproc`
```
b49020, line 88: Ditto.

.github/workflows/ci.yml
```diff
@@ -65,41 +65,110 @@ jobs:
...
-          make clean && make -j`nproc`
+          # Setup environment
+          export OPTEE_DIR=$HOME
+          export ARCH_TA=arm
+          export STD=y
+          source environment
 
+          # Build OP-TEE Rust examples for Arm 64-bit both host and TA
+          make -j`nproc`
+      - name: Run tests for Arm 64-bit both host and TA
```
b49020, line 165: I suppose you meant here 32-bit TAs instead.

b49020 (approved):
I just have some minor comments related to CI but the rest looks good to me.

.github/workflows/ci.yml
DemesneGH: done

b49020:
PR looks fine to me for a merge now.

DemesneGH:
merged, thanks!

Issue #148, opened 2024-11-14
If we can upgrade version of bitflags in optee-utee


ivila:
In [cargo.toml](https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/b5143c627de2ee128732fcf798c41516ffbc13f7/optee-utee/Cargo.toml#L30) of optee-utee crate, it lock the version of bitflags(set it must to 1.0.4):
![image](https://github.com/user-attachments/assets/32147a40-4306-45b3-982c-9c51d2a7f805)

However, in some other crates will require a newer version, for example, the [raw-cpuid](https://crates.io/crates/raw-cpuid) require version upper 1.2, and result in confliction:
![image](https://github.com/user-attachments/assets/9e2fbe33-2c4b-4efc-8748-fffb90df7256)
So I wonder why optee-utee lock the version of bitflags and if it can upgrade the version of bitflags

b49020:
I don't think upgrading dependencies should cause any issues, please feel free to propose corresponding PR.

PR #149, opened 2024-11-14
optee-utee: do not lock the version of bitflags crate

ivila:
Unlock the version of bitflags in optee-utee to avoid confliction when using optee-utee with some other crates(like cpuid).

This PR relate to issue #148


b49020:
This patch isn't specific to `no-std` as `optee-utee` is a common crate, so you can drop `no-std` from patch subject. With that addressed, feel free to add:

ivila:
Done, already drop the `no-std` prefix and add `Reviewed-by` and `Signed-off-by` in the commit message.

DemesneGH:
LGTM, merged, thanks!

Issue #151, opened 2024-11-28
Unexpected format result with Uuid ToString method


ivila:
## Problem

I found some unexpected results when using the ToString method of Uuid, In both optee_utee and optee_teec:
### 1. the output formation is different with input
the parse_str method require formation of neither **32 bytes or 36 bytes** (just what the [uuid](https://crates.io/crates/uuid) crate defined), however, the realization of Display trait of optee_teec::Uuid and optee_utee::Uuid will output a str with **35 bytes**.
for example:
```rust
fn test() {
    let uuid_str = "6bfaa402-50ed-42f2-9bde-b2aecfa216c2";
    let uuid = optee_teec::Uuid::parse_str(uuid_str).unwrap();
    let uuid_str2 = uuid.to_string();  // this is "6bfaa402-50ed-42f2-9bdeb2aecfa216c2", the last "-" is missing
}
```

### 2. missing the prefix zeros:
the formation teaclave use is `"{:x}-{:x}-{:x}-{}"`, which will get an unexpected result when format number has zeros at front, for example:
```rust
fn main() {
    let uuids = vec![
        "00173366-2aca-49bc-beb7-10c975e6131e",
        "11173366-0aca-49bc-beb7-10c975e6131e",
        "11173366-2aca-09bc-beb7-10c975e6131e",
        "11173366-2aca-19bc-beb7-10c975e6131e",
    ];
    for (i, raw) in uuids.iter().enumerate() {
        let tmp = optee_teec::Uuid::parse_str(raw).unwrap();
        let formatted = tmp.to_string();
        let raw2 = to_same_format(raw);
        println!(
            "str {}\norigin: {}\nparsed: {}\nequal: {}\n",
            i,
            raw2,
            formatted,
            raw2.eq_ignore_ascii_case(&formatted),
        );
    }
}

fn to_same_format(origin: &str) -> String {
    assert_eq!(origin.len(), 36);
    let mut buffer = Vec::with_capacity(35);
    buffer.extend(origin.as_bytes()[0..23].iter());
    buffer.extend(origin.as_bytes()[24..36].iter());
    assert_eq!(buffer.len(), 35);
    String::from_utf8(buffer).unwrap()
}
```
the codes will produce: 
```shell
str 0
origin: 00173366-2aca-49bc-beb710c975e6131e
parsed: 173366-2aca-49bc-beb710c975e6131e
equal: false

str 1
[...]

PR #152, opened 2024-11-29
Change the output formation of ToString method of Uuid

ivila:
1) change the realization of Display trait of Uuid, make its formation
   const of 36 bytes.
2) introduce an env(RUN_TEST_IN_REE) in optee-utee-sys crate so that
   we can run library tests of optee-utee

This one relates to #151

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,10 @@ jobs:
           export OPTEE_DIR=$HOME
           source environment
 
+          # Run unit tests
+          (cd optee-utee && SYS_BUILD_TYPE=unit_test cargo test --lib --features no_panic_handler -vv)
+          (cd optee-teec && SYS_BUILD_TYPE=unit_test cargo test --lib -vv)
+
           # Build Rust optee-utee and optee-teec
           (cd optee-utee && cargo build --target aarch64-unknown-linux-gnu -vv)
           (cd optee-teec && cargo build --target aarch64-unknown-linux-gnu -vv)
--- a/optee-teec/optee-teec-sys/build.rs
+++ b/optee-teec/optee-teec-sys/build.rs
@@ -19,6 +19,21 @@ use std::env;
 use std::path::Path;
 
 fn main() {
+    const ENV_SYS_BUILD_TYPE: &str = "SYS_BUILD_TYPE";
+    println!("cargo:rerun-if-env-changed={}", ENV_SYS_BUILD_TYPE);
+
+    let build_type = env::var(ENV_SYS_BUILD_TYPE).unwrap_or(String::from("")).to_lowercase();
+    match build_type.as_str() {
+        "unit_test" => unit_test_build(),
+        _ => production_build(),
+    }
+}
+
+// this allow developers to run unit tests in host machine(even x86)
+fn unit_test_build() {
+}
+
+fn production_build() {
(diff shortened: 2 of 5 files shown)


DemesneGH:

b49020:
@ivila care to update CI to run the library tests too?

ivila:
> @ivila care to update CI to run the library tests too?

@b49020 I have tried, the test of optee-utee can run succuessfully(because it links nothing when test), but for the test of optee-teec, it requires run on Arm machine (it requires to link the libteec, but on x86_64 machine it can not be done as we are not going to build a libteec.a of x86 architecture, check [this pipeline](https://github.com/ivila/incubator-teaclave-trustzone-sdk/actions/runs/12078291479/job/33682470731)).
image
And If I switch the pipeline to Arm64 runners I think it's a big change, shouldn't be include in this PR, that's why I finally give up.

DemesneGH:
@ivila Could you provide the cargo test command you use? Seems you have posted them on the original commit.

b49020:
@ivila agree I think compiling and running tests on arm64 runners makes better sense. However, we should be able to compile native for x86 too for testing purposes. However, we should accumulate all the library tests under:
- optee-teec/systest
- optee-utee/systest

b49020:
BTW, if you want to push actual fix separately, we should be able to merge it till we sort out how to run library tests.

ivila:
> @ivila Could you provide the cargo test command you use? Seems you have posted them on the original commit.

@DemesneGH Sure.
#### 1. For optee-utee
the command is: `(cd optee-utee && RUN_TEST_IN_REE=true cargo test --lib --features no_panic_handler --target aarch64-unknown-linux-gnu -vv)`
need to set the env of RUN_TEST_IN_REE to true, and use the features of no_panic_handler, this can run successfully on x86 runner.

#### 2. for optee-teec
the command is: (cd optee-teec && cargo test --lib --target aarch64-unknown-linux-gnu -vv)
this will failed as it wants to link teec (needs to run on Arm64 runner, and currently github not open this for OSS project, see this: [[Make ARM64 Windows and Linux runner images available for free](https://github.com/actions/runner-images/issues/10820#top)])

ivila:
> BTW, if you want to push actual fix separately, we should be able to merge it till we sort out how to run library tests.

@b49020 actually there is a quick fix way, just to provide an env to tell optee-teec-sys not to link the teec library (just like what I do at optee-utee-sys)

ivila:
> @ivila agree I think compiling and running tests on arm64 runners makes better sense. However, we should be able to compile native for x86 too for testing purposes. However, we should accumulate all the library tests under:
> 
> * optee-teec/systest
> * optee-utee/systest

I think `accumulate all the library tests` is not a good idea for library unit tests, because you have to expose everything you want to test, for example, I have a function that just use in a module and actually I don't want to expose it, I have to make it pub if I want to accumulate it.
image

b49020:
> I think accumulate all the library tests is not a good idea for library unit tests,

Sounds reasonable from unit testing point of view.

ivila:
@DemesneGH @b49020 
Codes updated, the changes are:
1) change the env from RUN_TEST_IN_REE(**true**|**false**) to BUILD_TYPE(**unit_test**|**any other value is default**) for better understanding
2) provide this env to optee-teec-sys too
3) add unit test to ci, under `build-utee-teec`, and the pipeline result just check [here](https://github.com/ivila/incubator-teaclave-trustzone-sdk/actions/runs/12081358187/job/33690301335)
image

b49020:

Issue #155, opened 2024-12-20
Add optee-utee-build crate


ivila:
## Reason
There are some reasons that makes us struggle with current build process:
### 1) ta_static.rs
This file was copy again and again, and it is meaningless for the TA, and TA need to write some configurations consts with specific names to combine with it, developers keep repeating themselves and make some new comers confusing, especially the `EXT_PROP_VALUE_1` and `EXT_PROP_VALUE_2`.

### 2) the linking process
There are bunch of codes copy again and again, some of the new comers confuse about the codes in the build.rs, why they are here, what did they do, what is the ta.lds, etc.

### 3) the upgrade process
Sometimes we need to update the codes, for example:
1. update build and link script as you do: [examples: polish linking script](https://github.com/apache/incubator-teaclave-trustzone-sdk/commit/a6b5cfbfbbe039aa17d954122f95c7d2b8fcf158#diff-c0cdd7b28f558bd417069b8e60ed35b70ac1cd01e68e3c0ba6c7311a5a444e22L42-R42)
2. update `#[no_mangle]` to `#[unsafe(no_mangle)]` as rust 2024 required
3. add some extra properties

Every time we do so, we need to update every crates in our repo, and must provide a detail description letting developers know which line to change, however developers expect they should just upgrade the version rather than modify the ta_static.rs, build.rs, main.rs again and again.

## Proposal
We should add a `optee-utee-build` crate, this crate can:
1. provide a TAConfig struct and generate user_ta_header.rs automatically: just like [`prost-build`](https://docs.rs/prost-build/latest/prost_build/) for `prost`, developers could just include the generated file in their `src/main.rs`, save them from ta.static.rs and configurations consts with specific names.
2. handle linking automatically:  generate ta.lds and link to it automatically, with other linking process.
3. provide a easy way for upgrade: every time we need to change something about the building process, we just upgrade this crate.

By using this crate, the `build.rs` in hello-world-rs change to:
```rust
// new build.rs
use proto;
use optee_utee_build::{TAConfig, RustEdition, Error};

fn main() -> Result<(), Error> {
    let config = TAConfig::new_standard("0.1", "This is a hello world example.", "Hello World TA");
    optee_utee_build::build(RustEdition::Before2024, proto::UUID, config)

}
```
And for reference, current codes:
[...]

PR #156, opened 2024-12-20
Add optee-utee-build crate

### LINKED ISSUE #135: Build issues on aarch64 Linux hosts

ivila:
1. add optee-utee-build crate to make building process simpler
2. use optee-utee-build to build hello_world example
3. fix pipeline

this PR relates to #155, and should resolves #135 too.


ivila:
> also add documentation as a separate commit
@b49020 May I ask what the documentation you points to, the README.md or the comments in rust codes?

b49020:
> > also add documentation as a separate commit
> 
> @b49020 May I ask what the documentation you points to, the README.md or the comments in rust codes?

I will suggest to create a dedicated document titled something like "Tutorial: Writing Rust TAs using optee-utee-build" under https://github.com/apache/incubator-teaclave-trustzone-sdk/tree/main/docs

DemesneGH:
> I suppose it's independent of this PR which is something @DemesneGH should be able to fix in the dockerfile with required dependencies for CI.

Sure, I'm working on that.

ivila:
I still have some changes to make and I am writing the document of how to use this crate, everything should be done in 3 days, so we can pause reviewing this PR and wait until I am finish with it.

ivila:
@b49020 @DemesneGH  PR fixed.

ivila:
Update to remove the confusing extra properties of `gp.ta.version` and `gp.ta.description`, as in this [commit](https://github.com/linaro-swg/optee_examples/commit/20e5a612e6303820c1e41faf8aa265404b28de82#diff-a0e1cb79ba65ecda7851beef61ad66d58e4cfdc1568b0cfd58118e6442a0ab28L27) [jforissier](https://github.com/linaro-swg/optee_examples/commits?author=jforissier) said they are just `misleading TA properties`:
image

DemesneGH:
> > I suppose it's independent of this PR which is something @DemesneGH should be able to fix in the dockerfile with required dependencies for CI.
> 
> Sure, I'm working on that.

The PR for CI fix: https://github.com/apache/incubator-teaclave-trustzone-sdk/pull/157.

ivila:
update to use latest ci

docs/writing-rust-tas-using-optee-utee-build.md
```diff
@@ -0,0 +1,183 @@
+---
+permalink: /trustzone-sdk-docs/writing-rust-tas-using-optee-utee-build.md
+---
+
```
DemesneGH, line 4: When finalizing this documentation, please rewrap all text in this document at 80 characters, this helps to further review and see changes.
Reference: https://thoughtbot.com/blog/wrap-existing-text-at-80-characters-in-vim

optee-utee-build/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[package]
+name = "optee-utee-build"
```
DemesneGH, line 19: Do we need a publish of this crate on `crates.io`? After this PR merged I can do that.

.licenserc.yaml
```diff
@@ -31,3 +31,4 @@ header:
     - 'DISCLAIMER-WIP'
     - '*.json'
     - 'examples/tls_server-rs/ta/test-ca/**'
+    - 'optee-utee-build/test_files/**'
```
DemesneGH, line 34: Could the license header be added to the `test_files`?

examples/hello_world-rs/ta/Makefile
```diff
@@ -20,7 +20,7 @@ UUID ?= $(shell cat "../uuid.txt")
 TARGET ?= aarch64-unknown-linux-gnu
 CROSS_COMPILE ?= aarch64-linux-gnu-
 OBJCOPY := $(CROSS_COMPILE)objcopy
-LINKER_CFG := target.$(TARGET).linker=\"$(CROSS_COMPILE)ld.bfd\"
+LINKER_CFG := target.$(TARGET).linker=\"$(CROSS_COMPILE)gcc\"
```
DemesneGH, line 23: Using the `ld.bfd` works as well, right?

optee-utee-build/src/linker.rs
```diff
@@ -0,0 +1,175 @@
...
+    fn auto_detect_linker_type() -> LinkerType {
+        const ENV_RUSTC_LINKER: &str = "RUSTC_LINKER";
+        println!("cargo:rerun-if-env-changed={}", ENV_RUSTC_LINKER);
+        match env::var(ENV_RUSTC_LINKER) {
+            Ok(ref linker_name) if linker_name.ends_with("ld") || linker_name.ends_with("ld.bfd") => {
+                return LinkerType::LD;
+            },
+            _ => {},
+        };
+        return LinkerType::CC;
```
DemesneGH, line 173: simplify to:
Ok(xxxx) => LinkerType::LD,
_ => LinkerType::CC

optee-utee-build/src/ta_config.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub fn new_standard(
+        uuid_str: &str,
+        ta_version: &str,
+        ta_description: &str,
+    ) -> Result<Self, Error> {
+        Ok(Self {
+            uuid: uuid_str.try_into()?,
+            ta_flags: 0,
+            ta_data_size: 1 * 1024 * 1024,
+            ta_stack_size: 2 * 1024,
```
DemesneGH, line 99: Keep the default values same as the which in optee_examples: https://github.com/linaro-swg/optee_examples/blob/master/hello_world/ta/user_ta_header_defines.h#L47-L50

optee-utee-build/src/ta_config.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub ta_description: String,
+    pub trace_level: i32,
+    pub trace_ext_prefix: String,
+    pub ta_framework_stack_size: u32,
+    pub ext_properties: Vec<Property>,
+}
+
+impl TAConfig {
+    /// generate a standard config by uuid, retrieve version and description from cargo.toml
+    pub fn new_standard_with_cargo_env(uuid_str: &str) -> Result<Self, Error> {
```
DemesneGH, line 82: 1. How about naming this to "default config"? 
`new_default_with_cargo_env`
`new_default`
2. Could we have some description of when we should use `new_standard_with_cargo_env`? Why we read these values `CARGO_PKG_VERSION` `CARGO_PKG_DESCRIPTION`?

optee-utee-build/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+version = "0.2.0"
+authors = ["Teaclave Contributors <dev@teaclave.apache.org>"]
+license = "Apache-2.0"
+repository = "https://github.com/apache/incubator-teaclave-trustzone-sdk.git"
+edition = "2018"
+description = "Build tool for TA"
+
```
DemesneGH, line 26: When finalizing, please run `cargo fmt` for this crate

examples/hello_world-rs/ta/Makefile
ivila: > Using the `ld.bfd` works as well, right?

Yes, it did on cross_compiling, but not working if you compile on arm host machine, as ld.bfd defaultly trying to link to libgcc_s.so instead of libgcc.a if not cross compile(and the argument is `-lgcc_s`).
image
and with `-Bstatic` provided, it tries to find `libgcc_s.a`, but this is not valid as it should be in the name of `libgcc.a`

optee-utee-build/Cargo.toml
ivila: Yes, thanks in advance.

optee-utee-build/src/ta_config.rs
ivila: > How about naming this to "default config"?

changed

> Could we have some description of when we should use new_standard_with_cargo_env? 

done

> Why we read these values CARGO_PKG_VERSION CARGO_PKG_DESCRIPTION

those values are provided by cargo, see [Environment variables Cargo sets for crates](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates)

examples/hello_world-rs/ta/Makefile
DemesneGH: As you mentioned above, it's really the good point for solving another building issue https://github.com/apache/incubator-teaclave-trustzone-sdk/issues/135. 

BTW Could you provide some short guidance on how to build examples on ARM host machine (and the choice of linker) at https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/main/README.md#setup-building-environment
Such as adding two sections: "building on ARM" , "cross compile on x86_64"? I think it's valuable for our developers.

ivila: > > Using the `ld.bfd` works as well, right?
> 
> Yes, it did on cross_compiling, but not working if you compile on arm host machine, as ld.bfd defaultly trying to link to libgcc_s.so instead of libgcc.a if not cross compile(and the argument is `-lgcc_s`). image and with `-Bstatic` provided, it tries to find `libgcc_s.a`, but this is not valid as it should be in the name of `libgcc.a`

@DemesneGH sorry for my previous misleading response, actually when using `ld.bfd` as linker, its argument is `-Bdynamic -lgcc_s`, the problem of `cannot find -lgcc_s` is that ld.bfd defaultly do not know how to find the libgcc_s.so in host machine (actually gcc has a bunch of undocument configs in it).

Here is how it happens:
1. we have a build.rs, so cargo will need to build a `build script binary` first, the target is the same as host machine, in x64 is **x86_64-unknown-linux-gnu**, in aarch64 is **aarch64-unknown-linux-gnu**.
2. we set linker config by  `target.aarch64-unknown-linux-gnu="aarch64-linux-gnu-ld.bfd"`, please keep in mind this change the linker of target **aarch64-unknown-linux-gnu**.
3. then run `make` to build ta, it tries to build a `build script binary` first and then use it to build our source codes, and things going bad starts from here. When building the `build script binary` it requires libunwind for backtrace, adds the requirement of `gcc_s` (yes, this is not added by our source codes but the `build.rs`), but we change the linker (and yes, the target of build script is **aarch64-unknown-linux-gnu** when your host is aarch64 host), so everything fuckup, you will get a link error as you cannot find libgcc_s.so(only gcc knows how to find it, for example, on my coolpi4B with ubuntu23, it is /lib/aarch64-linux-gnu/libgcc_s.so.1).

So my answer is, never use ld.bfd as default linker in if your target is the same as your host machine(aka not cross compiling)

> ### Why it works on x64 host?
> as what I described above, the target of build script is **x86_64-unknown-linux-gnu**, and the target of our source codes is **aarch64-unknown-linux-gnu**, we just change the linker of **aarch64-unknown-linux-gnu**, the build script is fine here.

ivila: > As you mentioned above, it's really the good point for solving another building issue #135.
> 
> BTW Could you provide some short guidance on how to build examples on ARM host machine (and the choice of linker) at https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/main/README.md#setup-building-environment Such as adding two sections: "building on ARM" , "cross compile on x86_64"? I think it's valuable for our developers.

Actually there is no difference between the two environments when developing (same commands, same linux packages, same environments), I think I should just add a notification about the linker problem so that developers will know more about the building process (especially for the C developers, they may not know that there are `build script binary` and `source code binary`?

optee-utee-build/src/linker.rs
```diff
@@ -0,0 +1,182 @@
...
+/// ```
+///
+pub struct Linker {
+    linker_type: LinkerType,
+}
+
+impl Linker {
+    pub fn new(linker_type: LinkerType) -> Self {
+        Self { linker_type }
+    }
```
b49020, line 75: nit: prefer space between functions here and below.

optee-utee-build/src/linker.rs
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        println!("cargo:rustc-link-search={}", out_dir.display());
+        println!("cargo:rerun-if-changed={}", ta_lds_file_path.display());
+        println!("cargo:rustc-link-arg=-T{}", ta_lds_file_path.display());
+        Ok(())
+    }
+
+    fn auto_detect_linker_type() -> LinkerType {
+        const ENV_RUSTC_LINKER: &str = "RUSTC_LINKER";
```
b49020, line 171: Do you expect auto-detection with `RUSTC_LINKER` works when the TA developer has to setup `--config $(LINKER_CFG)` manually? Is there a possibility that we can move that manual linker configuration to this `optee-utee-build` crate too?

examples/hello_world-rs/ta/Makefile
DemesneGH: How about keeping this modification and add a simple comment upon this:
```
-LINKER_CFG := target.$(TARGET).linker=\"$(CROSS_COMPILE)ld.bfd\"
+ # Configure the linker to use GCC, which works on both cross-compilation and ARM machines
+LINKER_CFG := target.$(TARGET).linker=\"$(CROSS_COMPILE)gcc\"
```
Since it will be added into each example's Makefile, I'm not sure if we need to explain the details such as `build script binary` and `source code binary` here. 
Would it be more reasonable to create a separate short doc for the building process in `docs/`?

b49020: I think it's more preferable for `optee_utee_build` crate to consume `CROSS_COMPILE` environment variable and setup linker configuration accordingly.

And regarding documentation update, I don't think we want developers to learn about how `optee_utee_build` crate figures out linker given the `CROSS_COMPILE` path. Instead the documentation should rather be just about letting the user know how to build Rust TAs in either cross-compiled environment (x86 host) or native environment (arm64 host). I think that will just boil down to the value of `CROSS_COMPILE` variable being empty ("") for native compilation and ("path-to-gcc-cross-compiler") for cross compilation.

optee-utee-build/src/ta_config.rs
```diff
@@ -0,0 +1,193 @@
...
+    pub ta_description: String,
+    pub trace_level: i32,
+    pub trace_ext_prefix: String,
+    pub ta_framework_stack_size: u32,
+    pub ext_properties: Vec<Property>,
+}
+
+impl TAConfig {
+    /// Generate a default config by uuid, this is a wrapper of `new_default`
+    /// function, it retrieves version and description from cargo.toml by
```
DemesneGH, line 82: "cargo.toml" => "your TA's Cargo.toml"

optee-utee-build/src/ta_config.rs
DemesneGH: I agree with this and it will simplify the configuration, we can read the `version` and `description` from `Cargo.toml` then we don't have to set the TA's constants like the previous:
```
const TA_VERSION: &[u8] = b"0.1\0";
const TA_DESCRIPTION: &[u8] = b"This is a hello world example.\0";
```

optee-utee-build/src/linker.rs
ivila: > Do you expect auto-detection with RUSTC_LINKER works when the TA developer has to setup `--config $(LINKER_CFG)` manually?

Actually the environment `RUSTC_LINKER` indicate what developers setup for linker, it would be empty if TA developer didn't setup `--config $(LINKER_CFG)`, so yes, it would work if TA developer has to setup `--config $(LINKER_CFG)`.

>  Is there a possibility that we can move that manual linker configuration to this optee-utee-build crate too?

No, currently cargo not allow the linker to be changed by build script.

optee-utee-build/src/linker.rs
ivila: I will leave some comments on it, or `cargo fmt` would remove the empty line.

examples/hello_world-rs/ta/Makefile
ivila: I agree with @b49020, let's just make things simpler, we just tell TA developers setting their `CROSS_COMPILE` env, and that's enough.

But currently cargo do not allow change linker from with build script, so it still need to be done by Makefile.

ivila: After further investigation, I found that `aarch64-linux-gnu-gcc` and `arm-linux-gnueabihf-gcc` exists on aarch64 hosts too, so I will just change `setup.sh` to install the right toolchains on aarch64 host, and keep all the steps all the same on different hosts (it's more platform-independent because, when using Rust, there shouldn't be any difference regardless of the host system developers on)

b49020: Okay, that's fine. Let's do following additionally to further clarify things:

1. Add comment as @DemesneGH described above:

> How about keeping this modification and add a simple comment upon this:
> 
> ```
> -LINKER_CFG := target.$(TARGET).linker=\"$(CROSS_COMPILE)ld.bfd\"
> + # Configure the linker to use GCC, which works on both cross-compilation and ARM machines
> +LINKER_CFG := target.$(TARGET).linker=\"$(CROSS_COMPILE)gcc\"
> ```

2. Update https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/main/README.md#platforms with a subsection to say the supported host machines as `x86-64` and `arm64`.

b49020:
Thanks @ivila for your efforts addressing review comments, the last one pending for me is [here](https://github.com/apache/incubator-teaclave-trustzone-sdk/pull/156#discussion_r1898288529). With that addressed, feel free to add:

examples/hello_world-rs/ta/Makefile
ivila: @b49020 How about add "**Currently, we support building on both aarch64 and amd64 host machines, and they share the same steps.**” under [Setup Building Environment](https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/main/README.md#setup-building-environment), I feel like the [Platforms section]( https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/main/README.md#platforms) more like **running and debugging** rather than **building**.

b49020: > "Currently, we support building on both aarch64 and amd64 host machines, and they share the same steps.” under [Setup Building Environment](https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/main/README.md#setup-building-environment),

Let's rather be specific regarding host machines as output of command `uname -m` which gives you `x86_64` and `aarch64`. I don't have strong preference for the section to put this under.

b49020:
Once this PR gets merged, let's make all the examples to use `optee-utee-build` crate too which can be a followup PR.

ivila:
Is there anything I miss? I think this PR is inactive for a long time.😐

docs/migrating-to-new-building-env.md
```diff
@@ -2,6 +2,10 @@
 permalink: /trustzone-sdk-docs/migrating-to-new-building-env.md
 ---
 
+> After optee-utee-build release, this doc is keeping for developers 
+who intend to know the detail of building process, we suggest use 
+optee-utee-build for building instead.
```
DemesneGH, line 7: Add the link to optee-utee-build doc

optee-utee-build/src/ta_config.rs
```diff
@@ -0,0 +1,193 @@
...
+/// let ta_config = TAConfig::new_default(
+///     UUID,
+///     "0.1.0",
+///     "hello world",
+/// )?.ta_stack_size(10 * 1024).ta_data_size(32 * 1024);
+/// # Ok(())
+/// # }
+/// ```
+#[derive(Debug, Clone)]
+pub struct TAConfig {
```
DemesneGH, line 67: Use UpperCamelCase "TaConfig" according to Rust's naming conventions. Reference: https://rust-lang.github.io/api-guidelines/naming.html

optee-utee-build/src/ta_config.rs
```diff
@@ -0,0 +1,193 @@
...
+/// // a base64 string value, must not append '\0' at last, we will add it for you.
+/// PropertyValue::BinaryBlock("c2RmYXNm".to_string());
+/// PropertyValue::U64(1);
+/// # Ok(())
+/// # }
+#[derive(Debug, Clone)]
+pub enum PropertyValue {
+    Bool(bool),
+    U32(u32),
+    UUID(uuid::Uuid),
```
DemesneGH, line 169: naming convention: UUID => Uuid

optee-utee-build/src/error.rs
```diff
@@ -0,0 +1,43 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#[derive(Debug)]
+pub enum Error {
+    Env(std::env::VarError),
+    IO(std::io::Error),
```
DemesneGH, line 21: IO => Io

optee-utee-build/src/error.rs
```diff
@@ -0,0 +1,43 @@
...
+    UUID(uuid::Error),
```
DemesneGH, line 22: UUID => Uuid

optee-utee-build/src/linker.rs
```diff
@@ -0,0 +1,190 @@
...
+/// The type of the linker, there are difference when using gcc/cc or ld/lld as  
+/// linker, For example, `--sort-section=alignment` parameter changes to  
+/// `-Wl,--sort-section=alignment` when using gcc as linker.
+///
+/// CC: gcc, cc, etc.
+/// LD: ld, lld, ld.bfd, ld.gold, etc.
+#[derive(Debug, Clone)]
+pub enum LinkerType {
+    CC,
+    LD,
```
DemesneGH, line 35: Cc,
Ld

DemesneGH (commented):
After some minor recommendations resolved please feel free to add: `Reviewed-by: Yuan Zhuang <yuanz@apache.org>`

ivila:
update to add reviewer

DemesneGH:
Thanks! Will merge after the CI passed.

DemesneGH:
@ivila CI failed because:
```
24-12-31T02:52:27.0302935Z    Compiling ta v0.2.0 (/__w/incubator-teaclave-trustzone-sdk/incubator-teaclave-trustzone-sdk/examples/hello_world-rs/ta)
2024-12-31T02:52:27.0773416Z error[E0432]: unresolved import `optee_utee_build::TAConfig`
2024-12-31T02:52:27.0774266Z   --> build.rs:19:24
2024-12-31T02:52:27.0774662Z    |
2024-12-31T02:52:27.0775144Z 19 | use optee_utee_build::{TAConfig, RustEdition, Error};
2024-12-31T02:52:27.0776717Z    |                        ^^^^^^^^
2024-12-31T02:52:27.0777049Z    |                        |
2024-12-31T02:52:27.0777428Z    |                        no `TAConfig` in the root
2024-12-31T02:52:27.0777908Z    |                        help: a similar name exists in the module: `TaConfig`
2024-12-31T02:52:27.0778270Z 
2024-12-31T02:52:27.0819894Z For more information about this error, try `rustc --explain E0432`.
2024-12-31T02:52:27.0909479Z error: could not compile `ta` (build script) due to 1 previous error
```

ivila:
> @ivila CI failed because:
> 
> ```
> 24-12-31T02:52:27.0302935Z    Compiling ta v0.2.0 (/__w/incubator-teaclave-trustzone-sdk/incubator-teaclave-trustzone-sdk/examples/hello_world-rs/ta)
> 2024-12-31T02:52:27.0773416Z error[E0432]: unresolved import `optee_utee_build::TAConfig`
> 2024-12-31T02:52:27.0774266Z   --> build.rs:19:24
> 2024-12-31T02:52:27.0774662Z    |
> 2024-12-31T02:52:27.0775144Z 19 | use optee_utee_build::{TAConfig, RustEdition, Error};
> 2024-12-31T02:52:27.0776717Z    |                        ^^^^^^^^
> 2024-12-31T02:52:27.0777049Z    |                        |
> 2024-12-31T02:52:27.0777428Z    |                        no `TAConfig` in the root
> 2024-12-31T02:52:27.0777908Z    |                        help: a similar name exists in the module: `TaConfig`
> 2024-12-31T02:52:27.0778270Z 
> 2024-12-31T02:52:27.0819894Z For more information about this error, try `rustc --explain E0432`.
> 2024-12-31T02:52:27.0909479Z error: could not compile `ta` (build script) due to 1 previous error
> ```

Sorry about that, I always miss the `example` part😂

ivila:
update to fix the `compilation error of hello-world example`

DemesneGH:
FYI the `optee-utee-build` crate has been published on https://crates.io/crates/optee-utee-build

PR #160, opened 2025-01-16
EXPERIMENT: use ubuntu-24.04-arm

msgilligan:
This is just a quick test. The right solution would probably be to use a matrix and do both arm and intel builds.


ivila:
Fixed by #184

PR #164, opened 2025-01-20
optee-utee: support no-std net.

### LINKED ISSUE #162: Proposal: optee-utee: support no-std net.
# Background

Currently, Teaclave has a net module provided by mssun (#31) four years ago, which can only be accessed in a std environment. However, during my development, I frequently perform HTTP requests to the REE (some services inside Kubernetes) from within the TEE. At present, I am using unsafe and optee_utee_sys directly. However, I believe it would be more elegant for Teaclave to officially support no_std TcpStream and hide the unsafe code and optee_utee_sys from developers.

# Proposal

The changes are quite small, so I'll just summarize them:
1. Introduced [SocketError](https://github.com/ivila/incubator-teaclave-trustzone-sdk/blob/8e90e14a643a44cc995d2741e72a258c4055b1d1/optee-utee/src/net/error.rs#L21): an enum that encompasses all the error values returned by the GP TEE Sockets API.
2. Introduced [Socket struct and SocketAdapter trait](https://github.com/ivila/incubator-teaclave-trustzone-sdk/blob/feat_noStdNet_zc/optee-utee/src/net/socket.rs): The Socket struct handles socket operations (open, send, recv), while SocketAdapter accommodates various implementations of the GP TEE Sockets API.
3. Introduced [TcpAdapter and UdpAdapter for OP-TEE](https://github.com/ivila/incubator-teaclave-trustzone-sdk/blob/feat_noStdNet_zc/optee-utee/src/net/optee.rs): using them to construct TcpStream and UdpSocket.

You can check [this commit at my fork](https://github.com/ivila/incubator-teaclave-trustzone-sdk/commit/8e90e14a643a44cc995d2741e72a258c4055b1d1) for details

ivila:
1. make optee_utee::net::{TcpStream, UdpSocket} support no-std
2. update tcp_client-rs and udp_socket-rs to support running at no-std

Fixes #162


examples/udp_socket-rs/ta/src/main.rs
```diff
@@ -52,17 +53,64 @@ fn destroy() {
...
-    let mut stream = UdpSocket::connect("127.0.0.1", 34254).unwrap();
-    stream.write_all(b"[TA]: Hello, Teaclave!").unwrap();
+#[cfg(not(target_os = "optee"))]
+fn udp_socket() -> Result<()> {
+    use alloc::string::String;
+    use alloc::vec::Vec;
+    use optee_utee::net::Setup;
+
+    let setup = Setup::new_v4("127.0.0.1", 34254)?;
+    let mut stream = UdpSocket::open(setup).map_err(|err| {
```
DemesneGH, line 68: Seems the API design aligns with the OP-TEE raw API. Is it possible to make it consistent with the `std` API similar to `std::net`?  then the TA code can remain mostly unchanged for both `std` and `no-std`?

ivila: Actually, we cannot. This is a pain point in Rust: `std::io::Error` doesn't live in core because it uses unwinding, meaning many things cannot be used in no_std, especially `std::io::{Read, Write}`. Therefore, we cannot make them the same, because in std, the return error type should be `std::io::Error`, but we cannot use it in no_std.

ivila: > Seems the API design aligns with the OP-TEE raw API. Is it possible to make it consistent with the `std` API similar to `std::net`? then the TA code can remain mostly unchanged for both `std` and `no-std`?

Actually, we cannot. This is a pain point in Rust: `std::io::Error` doesn't live in core because it uses unwinding, meaning many things cannot be used in no_std, especially `std::io::{Read, Write}`. Therefore, we cannot make them the same, because in std, the return error type should be `std::io::Error`, but we cannot use it in no_std.

DemesneGH: We don't need to implement `std::io::{Read, Write}` like what we do in `std`. How about making the API compatible? like:
```
// the no_std TcpStream:
pub struct TcpStream {
   ...
}
impl TcpStream {
   pub fn connect(address: &str, port: u16) -> Result<Self, SocketError> {...}
   pub fn read(&mut self, buf: &mut [u8]) -> Result<Self, SocketError> {...}
   pub fn write(&mut self, buf: &[u8]) -> Result<Self, SocketError> {...}
   ...
}
```

In examples we don't need to modify the main code:
```
// use different TcpStream
#[cfg(not(target_os = "optee"))]
use optee_utee::net::no_std::TcpStream;  // module name can be adjusted, just example here
#[cfg(target_os = "optee")]
use optee_utee::net::std::TcpStream;

// expose the same API for examples
let mut stream = TcpStream::connect(xxxx);
stream.write(xxx);
stream.read(xxx);
```

examples/udp_socket-rs/ta/src/main.rs
```diff
@@ -74,13 +122,33 @@ fn udp_socket() {
...
         }
     }
     trace_println!("{}", String::from_utf8_lossy(&response));
+    Ok(())
 }
 
 include!(concat!(env!("OUT_DIR"), "/user_ta_header.rs"));
+
+/// Workaround for those rustc bugs:
+/// * https://github.com/rust-lang/rust/issues/47493
```
DemesneGH, line 138: Have you tries this workaround in the issue? https://github.com/rust-lang/rust/issues/47493#issuecomment-606385776

examples/udp_socket-rs/ta/src/main.rs
ivila: Changes made, but I split them into three traits in case we only need the `connect` method at times and want to avoid conflicts with other `Read/Write` traits (for example, the `io::{Read, Write}` in `http_io`).

optee-utee/src/net/optee_no_std.rs
```diff
@@ -0,0 +1,150 @@
...
+///             Err(e) => return Err(e),
+///         }
+///     }
+///     if !buf.is_empty() {
+///         return Err(SocketError::Tee(ErrorKind::Generic));
+///     }
+///     Ok(())
+/// }
+///
+/// fn write_with_compact_trait(stream: &mut Stream, buf: &mut [u8]) -> Result<usize, SocketError> {
```
b49020, line 107: Typo s/write_/read_/

optee-utee/src/net/optee_no_std.rs
```diff
@@ -0,0 +1,150 @@
...
+///     TcpStream::open(setup)
+/// }
+///
+/// fn connect_with_compact_trait(host: &str, port: u16) -> Result<TcpStream, SocketError> {
+///     use optee_utee::net::StdCompactConnect;
+///
+///     TcpStream::connect_v4(host, port)
+/// }
+/// ```
+pub trait StdCompactConnect: Sized {
```
b49020, line 37: I suppose by name here you meant "Std compatible connect", right? If yes then we should rename it rather as s/StdCompactConnect/StdCompatConnect/. Same applies to other trait names as well.

optee-utee/src/net/optee_std.rs
```diff
@@ -0,0 +1,144 @@
...
+    ) -> std::io::Result<Self> {
+        let setup = Setup::new(address, port, ip_version)
+            .map_err(|_| std::io::Error::new(std::io::ErrorKind::Other, "Invalid address"))?;
+        Ok(Self::open(setup).map_err(|err| Into::<std::io::Error>::into(err))?)
+    }
+    pub fn connect_v4(address: &str, port: u16) -> std::io::Result<Self> {
+        Self::connect_with_ip_version(address, port, raw::TEE_ipSocket_ipVersion::TEE_IP_VERSION_4)
+    }
+    pub fn connect_v6(address: &str, port: u16) -> std::io::Result<Self> {
+        Self::connect_with_ip_version(address, port, raw::TEE_ipSocket_ipVersion::TEE_IP_VERSION_4)
```
b49020, line 39: Shouldn't here be `TEE_IP_VERSION_6` instead? It really looks like we need an example test-case for IP v6 testing.

optee-utee/src/net/optee_std.rs
```diff
@@ -0,0 +1,144 @@
...
```
b49020, line 81: Ditto here, need to be fixed for v6.

optee-utee/src/net/optee_std.rs
```diff
@@ -0,0 +1,144 @@
...
+            Ok(v) => Ok(v as usize),
+            Err(err) => Err(Into::<std::io::Error>::into(err)),
+        }
+    }
+    fn flush(&mut self) -> std::io::Result<()> {
+        Ok(())
+    }
+}
+
+impl Into<std::io::Error> for SocketError {
```
b49020, line 109: I don't see any current use-case of `std::io::Error` currently in example TAs but rather we use OP-TEE specific error traits. So let's drop it until there is a real user.

ivila: This simplifies the `try` operation in this file; otherwise, we would need to write many `map_err` calls. However, it's fine to remove it if necessary.

ivila: If we decide to keep it, I will add comment here for description.

b49020: Okay I see, we need it for mapping errors. Lets keep it.

optee-utee/src/net/optee_std.rs
ivila: Currently stuck in enable ipv6 support in qemu, any suggestion? @DemesneGH @b49020
I try binding a ipv6 address in host, result is:  
```shell
called `Result::unwrap()` on an `Err` value: Os { code: 97, kind: Uncategorized, message: "Address family not supported by protocol" }
```

I try connecting to `ipv6.google.com`, result is cannot make the connection.

ivila: After testing, I am 100% sure that the reason is modules missing(which result in ipv6 module not loaded).
image
Can we update the [image](https://nightlies.apache.org/teaclave/teaclave-trustzone-sdk/) we are using, fix the missing modules? I think update rootfs is enough.

DemesneGH: The image comes from OP-TEE repo buildroot outputs, with default building configuration. Could you provide the build configuration flags to build the ipv6 modules if you're also using buildroot? I'm also working on rebuilding the image.

b49020: Okay so the OP-TEE buildroot setup requires an update here. If you succeed with IPv6 testing then please push those changes to upstream OP-TEE buildroot setup too.

ivila: > The image comes from OP-TEE repo buildroot outputs, with default building configuration. Could you provide the build configuration flags to build the ipv6 modules if you're also using buildroot? I'm also working on rebuilding the image.

I believe IPv6 support is already enabled in the Linux kernel by default. The issue seems to be that the modules were not installed into the root filesystem (no `/lib/modules` inside `rootfs.cpio.uboot`). To fix this, we likely just need to run `make INSTALL_MOD_PATH=${PATH_TO_ROOTFS} modules_install` in the Linux kernel build process.

b49020: OP-TEE buildroot setup doesn't package any kernel modules but rather only relies of kernel built-in modules only. Give following change a try if it works for you:

```
build$ git diff
diff --git a/kconfigs/qemu.conf b/kconfigs/qemu.conf
index 418888c..b2a7e00 100644
--- a/kconfigs/qemu.conf
+++ b/kconfigs/qemu.conf
@@ -3,6 +3,7 @@ CONFIG_OPTEE=y
 ### Enable 9P VFS
 CONFIG_NET_9P=y
 CONFIG_NET_9P_VIRTIO=y
+CONFIG_IPV6=y
 CONFIG_9P_FS=y
 CONFIG_9P_FS_POSIX_ACL=y
 CONFIG_HW_RANDOM=y
```

ivila: > OP-TEE buildroot setup doesn't package any kernel modules but rather only relies of kernel built-in modules only. Give following change a try if it works for you:
> 
> ```
> build$ git diff
> diff --git a/kconfigs/qemu.conf b/kconfigs/qemu.conf
> index 418888c..b2a7e00 100644
> --- a/kconfigs/qemu.conf
> +++ b/kconfigs/qemu.conf
> @@ -3,6 +3,7 @@ CONFIG_OPTEE=y
>  ### Enable 9P VFS
>  CONFIG_NET_9P=y
>  CONFIG_NET_9P_VIRTIO=y
> +CONFIG_IPV6=y
>  CONFIG_9P_FS=y
>  CONFIG_9P_FS_POSIX_ACL=y
>  CONFIG_HW_RANDOM=y
> ```

Yes, it works, so just change kconfig with `CONFIG_IPV6=y` is enough @DemesneGH @b49020

DemesneGH: The `optee-qemuv8-4.2.0-ubuntu-24.04.tar.gz` image has been updated, please just retry the ci test for ipv6.

ivila: > The `optee-qemuv8-4.2.0-ubuntu-24.04.tar.gz` image has been updated, please just retry the ci test for ipv6.

OK, will try it.

ivila: @DemesneGH not working, the newer kernel still built without ipv6 support, you can just try `ip -6 addr`, if ipv6 is enabled, then result should not be empty.

Currently, test with kernel in `optee-qemuv8-4.2.0-ubuntu-24.04.tar.gz`, result is empty.
image

And test with my local built kernel, result is not empty.
image

DemesneGH: Sorry for the wrong image, new image re-uploaded.

ivila: > Sorry for the wrong image, new image re-uploaded.

It works! Thanks for your effort!

ivila:
Force update for ipv6 test and some comments. There are some more changes in those files:
1. `ci/ci.sh`: move `tcp_client-rs` and `udp_client-rs` out of STD only.
2. `examples/tcp_client-rs`: add ipv6 test.
3. `examples/udp_client-rs`: add ipv6 test.

ivila:
@DemesneGH @b49020 Any updates

b49020:
There is CI failure related to 64-bit std TAs, you should be able to look into it. However, FWIW:

ivila:
> There is CI failure related to 64-bit std TAs, you should be able to look into it. However, FWIW:
> 
> `Reviewed-by: Sumit Garg <sumit.garg@linaro.org>`

The error is `error: failed to get bitcode from object file for LTO (could not find requested section)`
image
It happens occasionally; simply retry, and the test will eventually pass.

PS: the CI is passed at my repositories, check [here](https://github.com/ivila/incubator-teaclave-trustzone-sdk/actions/runs/13191824930):
image

b49020:
> It happens occasionally; simply retry, and the test will eventually pass.

Okay, looks like it's related to custom OP-TEE toolchain target OS config option for std TAs. @DemesneGH can you take a look at this issue? We should be able to merge this PR while filing a corresponding issue such that we can keep track of it.

ivila:
Can we retry the failed test? I think it’s simply a lag that causes digest-rs to exceed the 5-second limit, resulting in a Ctrl-C (signal 2).

All the tests in the CI passed on my repository, check [here](https://github.com/ivila/incubator-teaclave-trustzone-sdk/actions/runs/13194518687)
image

DemesneGH:
> > It happens occasionally; simply retry, and the test will eventually pass.
> 
> Okay, looks like it's related to custom OP-TEE toolchain target OS config option for std TAs. @DemesneGH can you take a look at this issue? We should be able to merge this PR while filing a corresponding issue such that we can keep track of it.

This PR is merged and a new issue is created. https://github.com/apache/incubator-teaclave-trustzone-sdk/issues/166

optee-utee/src/net/optee_std.rs
ivila: > Okay so the OP-TEE buildroot setup requires an update here. If you succeed with IPv6 testing then please push those changes to upstream OP-TEE buildroot setup too.

@DemesneGH I think we forget this step😂

DemesneGH: Yes and an issue is opened for tracking this: https://github.com/apache/incubator-teaclave-trustzone-sdk/issues/174

Issue #165, opened 2025-01-23
[DISCUSSION] Release Version Number and Schedule


DemesneGH:
# [DISCUSSION] Release Version Number and Schedule  

Hi all,  

We are currently preparing for a new release and need more clarity about the release plan. We would greatly appreciate your thoughts and suggestions to help us establish a more concrete and reasonable release scheme.  

## 1. Upcoming Release Version Number  

Given the significant changes introduced since the previous release (v0.3.0-rc1), including breaking changes to the build environment, we are considering two options for the upcoming release version number:  

- **Option 1:** Release as `v0.3.0-rc2`, which means a normal upgrade, will later be finalized as `v0.3.0` after the Apache release process.   
- **Option 2:** Bump the version to `v1.0.0`, marking this release as a major milestone.  

Both choices have different considerations, taking `v0.3.0` as default, let's discuss about moving to `v1.0.0`.  

### YES because:  

- According to [Semantic Versioning](https://semver.org/), a major version should be incremented when there are incompatible API changes. Our build environment refactoring introduces significant breaking changes that are not compatible with the legacy setup.  
- This release includes more substantial changes compared to any previous release and potentially future releases in the near term (more details on the appended `Release Notes`). Labeling it as v1.0.0 highlights it as a major milestone rather than just an incremental upgrade.  
- Following OP-TEE's approach—where they bumped from 3.x to 4.x due to compatibility changes, including APIs and tools, we can align with their versioning strategy.

### NO because: 

- In the Rust ecosystem, a `v1.0.0` release often implies stability and production readiness. Many Rust crates (e.g., Rustls) maintain 0.x versions even with breaking changes to indicate ongoing development. Should we continue with minor version increments until we achieve broader adoption and maturity?  
- While the build environment has undergone significant changes, there haven't been major **API** modifications. 

## 2. Release Schedule  

OP-TEE follows a quarterly release schedule for minor versions and introduces major versions when there are incompatible API changes.  For major versions we've discussed on topic 1. In this part let's talk about normal releases (minor versions):

### What should be the release schedule?  

- **Align with OP-TEE's quarterly schedule?**  
  - As the recommended Rust SDK for OP-TEE, synchronizing our releases could enhance alignment and compatibility.  
- **Follow an independent schedule?**  
  - Since OP-TEE upgrades do not always directly impact the Rust SDK, we could adopt our own release schedule based on our development needs. The challenge lies in determining an effective and practical schedule.

[...]

PR #167, opened 2025-02-12
Bump version to 0.4.0

DemesneGH:
Since there have been code changes in OP-TEE crates, examples, and projects since the previous release, we are bumping all versions to 0.4.0.

### DIFF
--- a/examples/acipher-rs/host/Cargo.toml
+++ b/examples/acipher-rs/host/Cargo.toml
@@ -17,7 +17,7 @@
 
 [package]
 name = "acipher-rs"
-version = "0.2.0"
+version = "0.4.0"
 authors = ["Teaclave Contributors <dev@teaclave.apache.org>"]
 license = "Apache-2.0"
 repository = "https://github.com/apache/incubator-teaclave-trustzone-sdk.git"
--- a/examples/acipher-rs/proto/Cargo.toml
+++ b/examples/acipher-rs/proto/Cargo.toml
@@ -17,7 +17,7 @@
 
 [package]
 name = "proto"
-version = "0.2.0"
+version = "0.4.0"
 authors = ["Teaclave Contributors <dev@teaclave.apache.org>"]
 license = "Apache-2.0"
 repository = "https://github.com/apache/incubator-teaclave-trustzone-sdk.git"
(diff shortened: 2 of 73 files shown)


DemesneGH:
Note that the optee Rust crates version (and some examples) bumped from 0.2.0 to 0.4.0 which due to a pending release v0.3.0 for now. For unify the release process we all bump to 0.4.0 and will follow a formal release plan since 0.4.0.

ivila:
I suggest that in the future, we change some of the version strings to `edition.workspace = true` to save developers from having to update them one by one repeatedly.
For example, here is how [burn-rs](https://github.com/tracel-ai/burn/blob/1962fb17141249c04325dd09c89ded0edb49a10f/examples/custom-cubecl-kernel/Cargo.toml#L7C1-L7C8) do in their repo:
image

PS: maybe we should add `publish = false` in the examples too? See [[The publish field](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field)] for reference.

DemesneGH:
@ivila 
> I suggest that in the future, we change some of the version strings to `edition.workspace = true` to save developers from having to update them one by one repeatedly.

We have a discussion about versioning in: https://github.com/apache/incubator-teaclave-trustzone-sdk/issues/165#issuecomment-2646859519. Could you help to review that proposal? 
Using workspace really reduce the effort when bump version, can we have a more detailed plan on that discussion thread? Different with other Rust crates, we have more complex building environment, and `std`, `no-std`. Should all crates be in one workspace, or we manage examples in one workspace? I think more details should be clarified.

> PS: maybe we should add `publish = false` in the examples too? See [[The publish field](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field)] for reference.

When we publish crates on `crates.io`, the examples are not included in the package—only the crate's code is. While setting `publish = false` can serve as a clearer note, it doesn’t have any actual effect. We can add it in the PR of using workspace further.

-----
As for this PR, we'd like not to block the release process, so the idea of workspace can be discussed and may be included in the next release. Please help to review the current version bump and give the `Reviewed-by` tag if you're okay with it. Thanks!

ivila:
LGTM😀.
`Reviewed-by: Zehui Chen <ivila@apache.org>`

Issue #166, opened 2025-02-07
Linking error for std TAs


DemesneGH:
When setting LTO=true, this error occasionally occurs:
error: failed to get bitcode from object file for LTO (could not find requested section)

Reference: 
https://github.com/apache/incubator-teaclave-trustzone-sdk/pull/164#issuecomment-2642043217

ivila:
It happens at ARM64 too, it just happens randomly.

ivila:
Fixed by https://github.com/apache/incubator-teaclave-trustzone-sdk/commit/ece214bbde7ff0e520f2489aa231e10a5217145c

PR #170, opened 2025-02-26
examples & projects: simplify proto crate

ivila:
This PR aims to simplify the proto crate by making the following changes:

1. **Truncate `uuid.txt` Before Compilation**
   The `build.rs` script in the proto crate currently serves only one purpose: reading the UUID from uuid.txt and removing the extra newline at the end. Instead of handling this at build time, we can clean up the file beforehand using a simple shell command:
   ```shell
   truncate -s 36 uuid.txt
   ```
   This allows us to remove `build.rs` entirely and eliminate the `OUTDIR` environment variable from `lib.rs`, making the crate cleaner and simpler.

2. **Use `num_enum` for Enum Conversion**
   The manual implementation for converting between `Command` and `u32` can be simplified using `num_enum`, reducing boilerplate and improving maintainability.

For the detail changes, you can refer to the first commit of this PR.


DemesneGH:
For the `uuid.txt`:
I checked the code history, the `build.rs` was designed to generate the UUID if it doesn't exist and the generation functionality has been removed in https://github.com/apache/incubator-teaclave-trustzone-sdk/commit/7de996cdbc9f454fd41b18150f8a90bd033b2703.  
For the current `build.rs`, I agree to remove it. What would happen if we don't remove the extra newline of `uuid.txt`? Could our building script tolerate this case? If so, developers would not have to `truncate` manually.

For the `num_enum`:
It's a nice cleanup and looks good for me.

ivila:
> For the `uuid.txt`: I checked the code history, the `build.rs` was designed to generate the UUID if it doesn't exist and the generation functionality has been removed in [7de996c](https://github.com/apache/incubator-teaclave-trustzone-sdk/commit/7de996cdbc9f454fd41b18150f8a90bd033b2703). For the current `build.rs`, I agree to remove it. What would happen if we don't remove the extra newline of `uuid.txt`? Could our building script tolerate this case? If so, developers would not have to `truncate` manually.
> 
> For the `num_enum`: It's a nice cleanup and looks good for me.

If we don't remove the extra newline of `uuid.txt`, the `parse_str` of `Uuid` will report an error because we pass a string with 37 bytes.
``` rust
    // called `Result::unwrap()` on an `Err` value: InvalidLength { expected: Any([36, 32]), found: 37 }
    let uuid = Uuid::parse_str(proto::UUID).unwrap();
```

DemesneGH:
How about `pub const UUID: &str = &include_str!("../../uuid.txt").trim();`?

ivila:
> How about `pub const UUID: &str = &include_str!("../../uuid.txt").trim();`?

You can't do it as `trim` is not a const function.
image

examples/supp_plugin-rs/proto/src/lib.rs
```diff
@@ -16,40 +16,28 @@
...
-        match value {
-            0 => Command::Ping,
-            _ => Command::Unknown,
-        }
-    }
-}
-
-pub const TA_UUID: &str = &include_str!(concat!(env!("OUT_DIR"), "/ta_uuid.txt"));
+// You can use `truncate -s 36 uuid.txt` to remove the extra newline at the end 
+// of `uuid.txt`.
```
DemesneGH, line 30: prefer more details such as:
If `Uuid::parse_str()` returns an InvalidLength error, there may be an extra newline in your `uuid.txt` file. You can remove it by running `truncate -s 36 uuid.txt`.

DemesneGH:
I don't have a better solution at the moment, so let's just leave a note for developers to handle this when adding new TAs. If we find a better solution in the future, we can revisit and improve this.

After the tiny comment resolved please feel free to add:

ivila:
image
Litemap release a [new version](https://crates.io/crates/litemap) 8 hours ago and set the MSRV to 1.81: 
image

ivila:
The new std test is failed by: #166 
image

DemesneGH:
> The new std test is failed by: #166 image

This issue is often observed when building with multiple parallel jobs. Try setting the make -j`nproc` to make -j2 in two STD jobs as the workaround.

ivila:
> > The new std test is failed by: #166 image
> 
> This issue is often observed when building with multiple parallel jobs. Try setting the make -j`nproc` to make -j2 in two STD jobs as the workaround.

It's OK, the test success on my fork: https://github.com/ivila/incubator-teaclave-trustzone-sdk/actions/runs/13557743386/job/37895227716
image

@DemesneGH Can you review my changes?

DemesneGH (approved):
BTW please shorten the commit message title to fit on one line.

PR #176, opened 2025-03-25
optee_teec: extension: fix macro errors

ivila:
1. Fix usage of Rust types as return values in FFI functions.
    https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/cf0f662eec9213be32bf972888f422575b03bc7a/optee-teec/src/extension.rs#L23-L33
    Currently, the plugin's methods use Rust's `optee_teec::Result` as the return value for `FFI`, which is incorrect since Rust does not guarantee a stable ABI. We need to change them to `optee_teec_sys::TEEC_Result`(alias of u32) instead.

2. Standardize the function signature to require returning a Result.
    Similar to the original signatures of TEEC functions, we currently require the plugin's methods to return an error.
    * For `init`: from `fn()` to `fn() -> optee_teec::Result<()>`
    * For `invoke_command`: from `fn(&mut optee_teec::PluginParameters)` to `fn(&mut optee_teec::PluginParameters) -> optee_teec::Result<()>`

    And in fact, the current examples of `supp_plugin-rs` already return a Result, even though this contradicts their function signatures.
    image

3. Use inner functions in generated code to prevent unexpected early returns.
    Previous macros simply duplicated the code into the new function.
    https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/cf0f662eec9213be32bf972888f422575b03bc7a/optee-teec/macros/src/lib.rs#L60-L67
    Which is buggy if developers have some return statements inside.
    For example, if developer define a function:
    ```rust
    fn init() {
        if method() {
            return;
        }
        // some other codes
    }
    ```
    It will be rewrite to:
    ```rust
    #[no_mangle]
    pub fn _plugin_init() -> optee_teec::Result<()> {
        if method() {
            return;  // this line is a syntax error
        }
        // some other codes
        Ok(())
    }
    ```

* Fix example: supp_plugin-rs.

### DIFF
--- a/examples/supp_plugin-rs/ta/src/main.rs
+++ b/examples/supp_plugin-rs/ta/src/main.rs
@@ -20,11 +20,11 @@
 
 extern crate alloc;
 
+use optee_utee::LoadablePlugin;
 use optee_utee::{
     ta_close_session, ta_create, ta_destroy, ta_invoke_command, ta_open_session, trace_println,
 };
-use optee_utee::{Error, ErrorKind, Parameters, Result, Uuid};
-use optee_utee::{LoadablePlugin};
+use optee_utee::{ErrorKind, Parameters, Result, Uuid};
 use proto::{Command, PluginCommand, PLUGIN_SUBCMD_NULL, PLUGIN_UUID};
 
 #[ta_create]
@@ -52,26 +52,32 @@ fn destroy() {
 #[ta_invoke_command]
 fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
     trace_println!("[+] TA invoke command");
-    let mut p0 = unsafe { params.0.as_memref().unwrap() };
+    let mut p0 = unsafe { params.0.as_memref()? };
     let inbuf = p0.buffer().to_vec();
-    trace_println!("[+] TA received value {:?} then send to plugin", p0.buffer());
-    let uuid = Uuid::parse_str(PLUGIN_UUID).unwrap();
+    trace_println!(
+        "[+] TA received value {:?} then send to plugin",
+        p0.buffer()
+    );
+    let uuid = Uuid::parse_str(PLUGIN_UUID).map_err(|err| {
+        trace_println!("Invalid PluginUuid: {:?}", err);
+        ErrorKind::BadParameters
+    })?;
 
     match Command::from(cmd_id) {
         Command::Ping => {
-            let mut plugin = LoadablePlugin::new(&uuid);
-            let outbuf = plugin.invoke(
-                PluginCommand::Print as u32, 
-                PLUGIN_SUBCMD_NULL, 
-                &inbuf
-            ).unwrap();
+            let plugin = LoadablePlugin::new(&uuid);
+            let outbuf = plugin.invoke(PluginCommand::Print as u32, PLUGIN_SUBCMD_NULL, &inbuf)?;
 
-            trace_println!("[+] TA received out value {:?} outlen {:?}", outbuf, outbuf.len());
+            trace_println!(
+                "[+] TA received out value {:?} outlen {:?}",
+                outbuf,
+                outbuf.len()
+            );
             trace_println!("[+] TA call invoke_supp_plugin finished");
 
             Ok(())
         }
-        _ => Err(Error::new(ErrorKind::BadParameters)),
+        _ => Err(ErrorKind::BadParameters.into()),
     }
 }
 
--- a/optee-teec/macros/src/lib.rs
+++ b/optee-teec/macros/src/lib.rs
@@ -20,7 +20,7 @@
 extern crate proc_macro;
 
 use proc_macro::TokenStream;
-use quote::quote;
+use quote::{quote, ToTokens};
 use syn::parse_macro_input;
 use syn::spanned::Spanned;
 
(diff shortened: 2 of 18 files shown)


examples/supp_plugin-rs/ta/src/main.rs
```diff
@@ -54,19 +54,24 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
...
-            let mut plugin = LoadablePlugin::new(&uuid);
-            let outbuf = plugin.invoke(
-                PluginCommand::Print as u32, 
-                PLUGIN_SUBCMD_NULL, 
-                &inbuf
-            ).unwrap();
+            let plugin = LoadablePlugin::new(&uuid);
+            let outbuf = plugin
+                .invoke(PluginCommand::Print as u32, PLUGIN_SUBCMD_NULL, &inbuf)
+                .unwrap();
```
DemesneGH, line 68: It's a legacy coding style issue that we use `unwrap()` in many examples. Please help eliminate its usage here to prevent potential panics.

optee-teec/macros/src/lib.rs
```diff
@@ -34,38 +34,71 @@ pub fn plugin_init(_args: TokenStream, input: TokenStream) -> TokenStream {
...
         #[no_mangle]
-        pub fn _plugin_init() -> optee_teec::Result<()> {
-            #f_block
-            Ok(())
+        pub fn _plugin_init() -> u32 {
+            fn inner() -> optee_teec::Result<()> {
+                #f_block
+            }
+            match inner() {
+                Ok(()) => 0,
```
DemesneGH, line 70: Prefer `TEE_SUCCESS` instead of `0`.

optee-teec/macros/src/lib.rs
```diff
@@ -78,50 +111,64 @@ pub fn plugin_invoke(_args: TokenStream, input: TokenStream) -> TokenStream {
...
         #[no_mangle]
         pub fn _plugin_invoke(
             cmd: u32,
             sub_cmd: u32,
-            data: *mut c_char,
+            data: *mut core::ffi::c_char,
             in_len: u32,
             out_len: *mut u32
-        ) -> optee_teec::Result<()> {
+        ) -> u32 {
```
DemesneGH, line 156: u32 => raw::TEEC_Result?

optee-teec/macros/src/lib.rs
```diff
@@ -78,50 +111,64 @@ pub fn plugin_invoke(_args: TokenStream, input: TokenStream) -> TokenStream {
...
+            };
             let outslice = params.get_out_slice();
             unsafe {
                 *out_len = outslice.len() as u32;
                 std::ptr::copy(outslice.as_ptr(), data, outslice.len());
-            }
-
-            Ok(())
+            };
+            return 0;
```
DemesneGH, line 170: 0 => TEE_SUCCESS?

optee-teec/macros/src/lib.rs
```diff
@@ -78,50 +111,64 @@ pub fn plugin_invoke(_args: TokenStream, input: TokenStream) -> TokenStream {
     let f = parse_macro_input!(input as syn::ItemFn);
     let f_vis = &f.vis;
     let f_block = &f.block;
-    let f_decl = &f.decl;
-    let f_inputs = &f_decl.inputs;
+    let f_sig = &f.sig;
+    let f_inputs = &f_sig.inputs;
+    let return_type = check_return_type(&f);
 
+    let valid_return_type = return_type == ReturnType::TeecEmptyResult;
```
DemesneGH, line 118: If we only accept the return type as TeecEmptyResult, why we need to define the ReturnType::No and ReturnType::Other?
In line 129 we can just `&& check_return_type()` and 
in check_return_type(): 
syn::ReturnType::Type(xxx) => checks in line92-95; true
_ => false

ivila:
image

Network Error, the pipeline pass on my fork: https://github.com/ivila/incubator-teaclave-trustzone-sdk/actions/runs/14073623792

optee-teec/macros/src/lib.rs
```diff
@@ -34,38 +34,56 @@ pub fn plugin_init(_args: TokenStream, input: TokenStream) -> TokenStream {
...
         .to_compile_error()
         .into();
     }
 
     quote!(
-        #[no_mangle]
-        pub fn _plugin_init() -> optee_teec::Result<()> {
-            #f_block
-            Ok(())
+        pub fn _plugin_init() -> optee_teec::optee_teec_sys::TEEC_Result {
```
DemesneGH, line 62: follow the same naming as `use optee_teec_sys as raw;` `raw::TEEC_Result`?

ivila: We can't do that, as it would easily result in **duplicate import** errors. **Global imports** should never be used in generated code.
image

DemesneGH: How about 
```
// Re-export optee_teec_sys so developers don't have to add them to their cargo
// dependencies.
pub use optee_teec_sys as raw;
```

```
pub fn _plugin_init() -> optee_teec::raw::TEEC_Result {
```

DemesneGH: If we `pub use optee_teec_sys as raw;` , for consistency, other modules (e.g. `extension.rs`) in `optee-utee/src` can be updated, from `use optee_utee_sys as raw;` to `use crate::raw;`

DemesneGH:
Thanks!

DemesneGH:
Please feel free to rebase and merge once CI passed 😄

PR #177, opened 2025-04-01
optee-utee: Enable TA-to-TA Invocation

DemesneGH:
This commit provides the Rust API for GP API TEE_InvokeTACommand(), which enable the user TA calls another user TA or pseudo TA.

- Add abstraction of TeeParameters and TaSession;
- Add inter-ta example and test scripts.


optee-utee/src/ta_session.rs
```diff
@@ -0,0 +1,108 @@
...
+// under the License.
+
+use core::ptr;
+use optee_utee_sys as raw;
+
+use crate::{Error, Result, TeeParam, TeeParameters, Uuid};
+
+/// Represents a connection between a trusted application and another trusted application (can be user TA or pseudo TA).
+pub struct TaSession {
+    raw: raw::TEE_TASessionHandle,
```
m4sterchain, line 25: We can maintain the timeout configuration in TaSession, to simplify the invoke_command interface.
We can further provide additional constructors with default timeout for a cleaner session create/invocation workflow.

optee-utee/src/tee_parameter.rs
```diff
@@ -0,0 +1,223 @@
...
+use optee_utee_sys as raw;
+
+pub trait TeeParam {
+    fn into_raw(&mut self) -> raw::TEE_Param;
+    fn param_type(&self) -> ParamType;
+    fn from_raw(raw: raw::TEE_Param, param_type: ParamType) -> Self;
+}
+
+pub struct TeeParameters<A, B, C, D> {
+    raw: [raw::TEE_Param; 4],
```
m4sterchain, line 29: I don't get the benefit of the design of `TeeParameters` here, especially compared with a basic Rust struct declaration with four fields (arg0, arg1, arg2, arg3).

DemesneGH: I've tried multiple designs, and my thought process is as follows:  

The first version:  
```rust
pub struct TeeParameters(p0: Parameter, p1: Parameter, p2: Parameter, p3: Parameter);
```
Here, `Parameter` is a union that can be one of three types: `TeeParamMemref`, `TeeParamValue`, or `TeeParamNone`.  
```rust
impl From<TeeParamMemref> for Parameter { ... }
```
However, when using this approach, we must explicitly call `.into()` like this:  
```rust
let mut parameters = Parameters(param_in.into(), param_out.into(), ParamNone.into(), ParamNone.into()); 
```
This `.into()` usage is not user-friendly, so I refined it in the second version:  

```rust
pub struct TeeParameters<A, B, C, D> {
    p0: A,
    p1: B,
    p2: C,
    p3: D,
}
pub trait TeeParam { ... }
impl TeeParam for TeeParamValue {...}
impl TeeParam for TeeParamMemref {...}
impl TeeParam for TeeParamNone {...}
```
To pass the parameters as an array to the raw API, I dynamically allocate the array. The pseudo-code is:  
```rust
impl<A: TeeParam, B: TeeParam, C: TeeParam, D: TeeParam> TeeParameters<A, B, C, D> {
    pub fn as_array(&self) -> [TEE_Param; 4] {
        // Convert each parameter to its raw representation
        [raw_p0, raw_p1, raw_p2, raw_p3]
    }
}

To invoke raw API:
raw::TEE_InvokeTACommand(params: params.as_array().as_mut_ptr(), ...)
```
However, this design has a flaw: we don't retain the raw array `[TEE_Param; 4]` address (which is passed directly to the raw API), so we cannot retrieve the output values.  

For example, in `hello_world`, the [TEE_Param; 4] memory layout looks like this:  
```
[
p0_inout_value_a, p0_inout_value_b,  // If a parameter is of type Value, it has two fields (a, b)
 none,  // p1
 none,  // p2
 none. // p3
]
```
When the TA writes back to `p0_inout_value_a`, we must maintain the array address to properly fetch the updated value.  

Then comes to the current definition, which is similar to existing `optee-teec/operation`:  
```rust
pub struct TeeParameters<A, B, C, D> {
    raw: [raw::TEE_Param; 4],  // Maintains the raw buffer for interaction with raw API
    param_types: ParamTypes,
    phantom0: PhantomData<A>,
    phantom1: PhantomData<B>,
    phantom2: PhantomData<C>,
    phantom3: PhantomData<D>,
}
```
This structure ensures that we correctly track and interact with the parameter buffer, and use trait to eliminate `into()`.

optee-utee/src/ta_session.rs
DemesneGH: Agree. Will update

examples/inter_ta-rs/ta/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+    ta_close_session, ta_create, ta_destroy, ta_invoke_command, ta_open_session, trace_println,
+};
+use optee_utee::{Error, ErrorKind, Parameters, Result, Uuid};
+use optee_utee::{TaSession, TeeParamMemref, TeeParamNone, TeeParamValue, TeeParameters};
+use optee_utee_sys as raw;
+use proto::Command;
+
+const SYSTEM_PTA_UUID: &str = "3a2f8978-5dc0-11e8-9c2d-fa7ae01bbebc";
+const SYSTEM_PTA_CMD_DERIVE_TA_UNIQUE_KEY: u32 = 1;
+const HELLO_WORLD_USER_TA_UUID: &str = "133af0ca-bdab-11eb-9130-43bf7873bf67";
```
ivila, line 31: Can we import the `proto` crate of `hello_world` instead of define them here?

PS: [renaming-dependency](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#renaming-dependencies-in-cargotoml)

examples/inter_ta-rs/ta/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+    let system_pta_uuid =
+        Uuid::parse_str(SYSTEM_PTA_UUID).map_err(|_| Error::from(ErrorKind::BadParameters))?;
+    let timeout = raw::TEE_TIMEOUT_INFINITE;
+    let mut session = TaSession::new(system_pta_uuid, timeout)?;
+    trace_println!("[+] TA open PTA session success");
+
+    let mut input: [u8; 32] = [0; 32];
+    let mut output: [u8; 32] = [0; 32];
+
+    let param_in = TeeParamMemref::new_input(&mut input);
```
ivila, line 66: Input might not be `mut` here.

examples/inter_ta-rs/ta/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    let mut parameters = TeeParameters::new(param_in, param_out, TeeParamNone, TeeParamNone);
+
+    session.invoke_command(
+        timeout,
+        SYSTEM_PTA_CMD_DERIVE_TA_UNIQUE_KEY,
+        &mut parameters,
+    )?;
+
+    let output = parameters.parameters().1.buffer().to_vec();
```
ivila, line 77: Do we need to convert them to vectors, or can we just use slice comparison here?

Or even further, just `slice.iter().all(|&x| x == 0)` to check if all zeroes in the slice.

examples/inter_ta-rs/ta/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+    trace_println!("[+] TA invoke command");
+    match Command::from(cmd_id) {
+        Command::Test => {
+            test_invoke_system_pta()?;
+            test_invoke_hello_world_user_ta()?;
+            trace_println!("[+] Test passed");
+            Ok(())
+        }
+        _ => {
+            return Err(Error::new(ErrorKind::NotSupported));
```
ivila, line 118: Can remove the `return` keyword here.

optee-utee/src/ta_session.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+impl TaSession {
+    /// Initializes a new TA Session.
+    pub fn new(uuid: Uuid, timeout: u32) -> Result<Self> {
+        // let mut raw_session = raw::TEE_HANDLE_NULL;
+        let mut raw_session: raw::TEE_TASessionHandle = ptr::null_mut();
+        let mut err_origin: u32 = 0;
+        unsafe {
```
ivila, line 34: I think `match unsafe { f() } { ...... }` is better.

optee-utee/src/ta_session.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    /// Initializes a new TA Session with parameters.
+    pub fn new_with_params<A: TeeParam, B: TeeParam, C: TeeParam, D: TeeParam>(
+        uuid: Uuid,
+        timeout: u32,
+        params: &mut TeeParameters<A, B, C, D>,
+    ) -> Result<Self> {
+        let mut raw_session: raw::TEE_TASessionHandle = ptr::null_mut();
+        let mut err_origin: u32 = 0;
+        unsafe {
```
ivila, line 57: Same as line 34

optee-utee/src/ta_session.rs
```diff
@@ -0,0 +1,108 @@
...
+                &mut err_origin,
+            ) {
+                raw::TEE_SUCCESS => Ok(Self { raw: raw_session }),
+                code => Err(Error::from_raw_error(code)),
+            }
+        }
+    }
+
+    /// Converts a TA Session to a raw pointer.
+    pub fn as_mut_raw_ptr(&mut self) -> *mut raw::TEE_TASessionHandle {
```
ivila, line 73: Can we just make it `pub(crate)`?

optee-utee/src/ta_session.rs
ivila: Maybe in `TeeParameters`?

examples/inter_ta-rs/ta/build.rs
```diff
@@ -0,0 +1,24 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use optee_utee_build::{Error, RustEdition, TaConfig};
+use proto;
```
ivila, line 19: Can remove this import.

optee-utee/src/tee_parameter.rs
```diff
@@ -0,0 +1,223 @@
...
+            size: buffer.len() as usize,
+        };
+        Self {
+            raw,
+            param_type,
+            _marker: PhantomData,
+        }
+    }
+
+    pub fn new_input(buffer: &'parameter mut [u8]) -> Self {
```
ivila, line 165: Input should not be mut here.

examples/inter_ta-rs/ta/src/main.rs
DemesneGH: Since we just need the uuid and cmd, import hello-world-rs seems make dependency complicated, we cannot build the inter-ta example without hello-world example. Another consideration is, system PTA is not Rust example, it's a C implementation in OS, we cannot import it. How about we add the definition in inter_ta-rs/proto/src/lib.rs:
```
pub enum Command {
    Test,
    #[default]
    Unknown,
}

pub enum HelloWorldTaCommand {
    IncValue,
    DecValue,
    #[default]
    Unknown,
}

pub enum SystemPtaCommand {
    DeriveTaUniqueKey,
    ...,
    #[default]
    Unknown,
}

pub const UUID: &str = &include_str!("../../uuid.txt");
pub const SYSTEM_PTA_UUID: &str = "3a2f8978-5dc0-11e8-9c2d-fa7ae01bbebc";
pub const HELLO_WORLD_USER_TA_UUID: &str = "133af0ca-bdab-11eb-9130-43bf7873bf67";

```

ivila: Agree😄

DemesneGH:
Closing as resolved in https://github.com/apache/incubator-teaclave-trustzone-sdk/pull/178

PR #178, opened 2025-04-08
New Design for TA-to-TA Invocation

DemesneGH:
Compared to https://github.com/apache/incubator-teaclave-trustzone-sdk/pull/177, the Params and TaSession are re-organized and can be used in a more Rust style.
It also addressed some comments from the  previous PR https://github.com/apache/incubator-teaclave-trustzone-sdk/pull/177.


optee-utee/src/tee_parameter.rs
```diff
@@ -0,0 +1,302 @@
...
+            ParamContent::ValueOutput { a, b } => raw::TEE_Param {
+                value: raw::Value { a: *a, b: *b },
+            },
+        }
+    }
+
+    fn update_size_from_raw(&mut self, raw_param: &raw::TEE_Param) {
+        match &mut self.content {
+            ParamContent::MemrefOutput { buffer: _, written } => {
+                *written = unsafe { raw_param.memref.size };
```
m4sterchain, line 152: I am curious why we consider accessing `raw::TEE_Param` is `unsafe` here. Can you add comment about the reason to the code?

optee-utee/src/tee_parameter.rs
```diff
@@ -0,0 +1,302 @@
...
+            }
+            _ => {}
+        }
+    }
+
+    fn update_value_from_raw(&mut self, raw_param: &raw::TEE_Param) {
+        match &mut self.content {
+            ParamContent::ValueInout { a, b } => {
+                *a = unsafe { raw_param.value.a };
```
m4sterchain, line 164: I am curious why we consider accessing `raw::TEE_Param` is `unsafe` here. Can you add comment about the reason to the code?

m4sterchain (commented):
The abstraction looks clear and well-structured. I've added a few minor comments for consideration. Great work overall!

optee-utee/src/tee_parameter.rs
```diff
@@ -0,0 +1,302 @@
...
+        }
+    }
+
+    fn update_size_from_raw(&mut self, raw_param: &raw::TEE_Param) {
+        match &mut self.content {
+            ParamContent::MemrefOutput { buffer: _, written } => {
+                *written = unsafe { raw_param.memref.size };
+            }
+            ParamContent::MemrefInout { buffer: _, written } => {
+                *written = unsafe { raw_param.memref.size };
```
m4sterchain, line 155: Should we check against the out `size` before assign to `written`? This function could be improved to return Result<()>.

optee-utee/src/tee_parameter.rs
```diff
@@ -0,0 +1,302 @@
...
+        for (i, param) in self.params.iter_mut().enumerate() {
+            let raw_param = &raw_params[i];
+            match param.get_type() {
+                ParamType::MemrefOutput => {
+                    param.update_size_from_raw(raw_param);
+                }
+                ParamType::MemrefInout => {
+                    param.update_size_from_raw(raw_param);
+                }
+                ParamType::ValueInput => {
```
m4sterchain, line 277: This should be ValueOutput.

optee-utee/src/ta_session.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    /// Set the default timeout for this session (mutable version).
+    /// Returns self to allow method chaining with invoke().
+    pub fn with_timeout(&mut self, timeout: u32) -> &mut Self {
+        self.default_timeout = timeout;
+        self
+    }
+
+    /// Open the session without parameters.
+    pub fn open(mut self) -> Result<Self> {
```
ivila, line 55: I think we should remove the `new` method and make `open` and `open_with_params` static methods instead. The current new method returns an invalid instance, and you must call `open` or `open_with_params` before doing anything useful—so it’s cleaner to just drop new altogether.

optee-utee/src/ta_session.rs
```diff
@@ -0,0 +1,150 @@
...
+    /// Can be called both before and after opening the session.
+    pub fn timeout(mut self, timeout: u32) -> Self {
+        self.default_timeout = timeout;
+        self
+    }
+
+    /// Set the default timeout for this session (mutable version).
+    /// Returns self to allow method chaining with invoke().
+    pub fn with_timeout(&mut self, timeout: u32) -> &mut Self {
+        self.default_timeout = timeout;
```
ivila, line 50: Should keep the same naming strategy with other files(like `tee_parameter.rs`).
1. `pub fn timeout(mut self)` => `pub fn set_timeout(&mut self)`
    prefix with `set_`, and change `mut self` to `&mut self`
3. `pub fn with_timeout(&mut self)` => `pub fn with_timeout(mut self)`
    change `&mut self` to `mut self`

optee-utee/src/ta_session.rs
```diff
@@ -0,0 +1,150 @@
...
+                self.raw = raw_session;
+                Ok(self)
+            }
+            code => Err(Error::from_raw_error(code)),
+        }
+    }
+
+    /// Open the session with parameters.
+    pub fn open_with_params(mut self, params: &mut TeeParams) -> Result<Self> {
+        let mut err_origin: u32 = 0;
```
ivila, line 79: The `err_origin` never return, but I think some of the developers might read it (same as the `ERROR_TARGET_DEAD` problem in teec).

optee-utee/src/ta_session.rs
```diff
@@ -0,0 +1,150 @@
...
+                // Update the parameters with the results
+                params.update_from_raw(&raw_params);
+                trace_println!("TEE_InvokeTACommand: command_id: {} finished", command_id);
+                Ok(())
+            }
+            code => Err(Error::from_raw_error(code)),
+        }
+    }
+
+    pub fn as_mut_raw_ptr(&mut self) -> *mut raw::TEE_TASessionHandle {
```
ivila, line 136: Shouldn't exposed to developers, maybe `pub(crate)`?

optee-utee/src/ta_session.rs
```diff
@@ -0,0 +1,150 @@
...
+        &mut self.raw
+    }
+}
+
+// Drop implementation to close the session
+impl Drop for TaSession {
+    fn drop(&mut self) {
+        if !self.raw.is_null() {
```
ivila, line 144: By removing the `new` method, we wouldn't need to check if `self.raw` is valid any more.

optee-utee/src/tee_parameter.rs
```diff
@@ -0,0 +1,302 @@
...
+}
+
+impl<'a> Param<'a> {
+    fn new() -> Self {
+        Self {
+            content: ParamContent::None,
+        }
+    }
+
+    pub fn written_slice(&self) -> Option<&[u8]> {
```
ivila, line 79: Maybe return a `Result` type with `TEE_ERROR_BAD_FORMAT` or `TEE_ERROR_BAD_PARAMETERS`, and developers can easily convert the `Result` to `Option` using `ok()` or `err()` method.

optee-utee/src/ta_session.rs
DemesneGH: I want to follow the builder's pattern, the configuration it can take is `timeout`. We can use default timeout:
```
let mut session = TaSession::new(system_pta_uuid).open()?;
```
Or set a timeout before open:
```
let mut session = TaSession::new(system_pta_uuid).timeout(1000).open()?;
```
For invoke we can overwrite the timeout:
```
let mut session = TaSession::new(system_pta_uuid).open()?;
session.with_timeout(2000).invoke_command(xxxx)
```
Maybe I should add some comments and one example usage in this inter_ta-rs/ta

optee-utee/src/tee_parameter.rs
DemesneGH: Accessing `union` should be unsafe in Rust:
error[E0133]: access to union field is unsafe and requires unsafe function or block
   --> /teaclave/optee-utee/src/tee_parameter.rs:152:28
    |
152 |                 *written = raw_param.memref.size;
    |                            ^^^^^^^^^^^^^^^^ access to union field
    |
    = note: the field may not be properly initialized: using uninitialized data will cause undefined behavior

optee-utee/src/tee_parameter.rs
```diff
@@ -0,0 +1,302 @@
...
+                ParamType::MemrefInout => {
+                    param.update_size_from_raw(raw_param);
+                }
+                ParamType::ValueInput => {
+                    param.update_value_from_raw(raw_param);
+                }
+                ParamType::ValueInout => {
+                    param.update_value_from_raw(raw_param);
+                }
+                _ => {}
```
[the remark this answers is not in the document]
DemesneGH: > 1. `ValueInput` contents are never changed; if not, raise an error;

For ValueInput, we will not read the TEEParam buffer after invoke, which means if it was changed we will ignore it. It is the same strategy as MemrefInput. Should we add the check for those input?

For `2.`, agree and will update

optee-utee/src/ta_session.rs
DemesneGH: It is not used, removed the function.

optee-utee/src/tee_parameter.rs
m4sterchain: I suggest we follow `https://github.com/rust-lang/rust-clippy/issues/9330` to explicitly add comment to `unsafe` block; and run `cargo lint` to help automatically enforce available rules.

optee-utee/src/tee_parameter.rs
DemesneGH: > developers can easily convert the `Result` to `Option` using `ok()` or `err()` method.

Yes it also works but follows a different design choice.

In current strategy, we choose to return an Option to indicate whether a written slice is available, allowing developers to decide whether to raise an error.

optee-utee/src/ta_session.rs
DemesneGH: Agree, will follow the definition in optee-teec:
```
pub struct Error {
    kind: ErrorKind,
    origin: Option<ErrorOrigin>,
}
```

optee-utee/src/tee_parameter.rs
DemesneGH: Will add `SAFETY` comment for the modules inside this PR. Since `cargo clippy` infects many other modules, there will be another PR to address all rules.

optee-utee/src/ta_session.rs
ivila: Maybe consider introducing a `TaSessionBuilder` struct? Mixing them together isn't very ergonomic.

Also, maybe consider supporting timeout configuration in `TeeParams`. That way, developers can use `open_with_params` to set the timeout, and we wouldn't need a separate timeout argument for `open`.

optee-utee/src/tee_parameter.rs
ivila: Got it, in that case, could you please add some comments explaining the design at the top of the Param struct?

optee-utee/src/ta_session.rs
DemesneGH: > Also, maybe consider supporting timeout configuration in `TeeParams`.

If a developer wants to configure both parameters and timeout, it makes sense to include the timeout in TeeParams and call `open_with_params(params)`.
But if a developer only wants to set a timeout, requiring them to create a TeeParams just for that feels unintuitive.

I've updated a more clear abstraction for TaSessionBuilder.

optee-utee/src/tee_parameter.rs
DemesneGH: done

optee-utee/src/tee_parameter.rs
```diff
@@ -0,0 +1,302 @@
...
+impl<'a> TeeParams<'a> {
+    pub fn new() -> Self {
+        Self {
+            params: [Param::new(), Param::new(), Param::new(), Param::new()],
+        }
+    }
+
+    // Method-chaining parameter setters
+    pub fn with_memref_in(mut self, idx: ParamIndex, buffer: &'a [u8]) -> Self {
+        self.params[idx.to_usize()].content = ParamContent::MemrefInput { buffer };
```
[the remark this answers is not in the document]
DemesneGH: done

ivila:
`Reviewed-by: Zehui Chen <ivila@apache.org>`

ivila:
Should we merge this PR or are we waiting for something?

DemesneGH:
Any other comments? @m4sterchain

m4sterchain:
LGTM.

PR #179, opened 2025-04-14
Fix teec ctest zc

### LINKED ISSUE #168: Systest broken
I tried `cargo build` in [optee-utee/systest](https://github.com/apache/incubator-teaclave-trustzone-sdk/tree/main/optee-utee/systest), and it ends up with some compilation errors:

Image
Additionally,  I found that [ctest](https://crates.io/crates/ctest) has been unmaintained for years and there is a fork [ctest2](https://crates.io/crates/ctest2) 

After making the following modifications, I finally got it to run:
* Changed `ctest@0.2` to `ctest2@0.4`
* Updated the include path from `env::var("TA_DEV_KIT_DIR").unwrap()` to `Path::new(env::var("TA_DEV_KIT_DIR").unwrap().as_str()).join("include")`

I encountered a bunch of errors, and the strangest one is that it generates a weird type called `signed ize_t` in the output file `all.c`.
I'm quite sure it should be `size_t`, so I suspect there are bugs in `ctest2`, but I haven't investigated further.

Image

Does anyone know how to fix this? Or should we just leave it unfixed and remove it?

ivila:
This PR fix the systest of optee-teec and optee-utee, fixes #168

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,8 @@ defaults:
     shell: bash
 
 jobs:
+  # Cross-compile for ARM64 on an AMD64 host and run unit-tests of (optee-utee
+  # and optee-teec) on AMD64 host.
   build-utee-teec:
     runs-on: ubuntu-latest
     container: yuanz0/teaclave-trustzone-sdk:ubuntu-24.04
@@ -50,6 +52,15 @@ jobs:
           (cd optee-utee && cargo build --target aarch64-unknown-linux-gnu -vv)
           (cd optee-teec && cargo build --target aarch64-unknown-linux-gnu -vv)
 
+  # Cross-compile on AMD64 and run tests in QEMU
+  #
+  # Cross-compile target pairs:
+  # - (arm32 host, arm32 ta)
+  # - (arm32 host, arm64 ta)
+  # - (arm64 host, arm32 ta)
+  # - (arm64 host, arm64 ta)
+  #
+  # Run tests target: (arm64 host, arm64 ta)
   build-and-test-examples:
     runs-on: ubuntu-latest
     container: yuanz0/teaclave-trustzone-sdk:ubuntu-24.04
@@ -93,6 +104,8 @@ jobs:
         run: |
           source environment
           (cd ci && ./ci.sh)
+
+  # Cross-compile for ARM64 on AMD64 and run tests in QEMU
   build-and-test-examples-for-64bit-std-TAs:
     runs-on: ubuntu-latest
     container: yuanz0/teaclave-trustzone-sdk:ubuntu-24.04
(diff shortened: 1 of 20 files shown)


.github/workflows/ci.yml
```diff
@@ -211,6 +211,28 @@ jobs:
           make -j3 toolchains &&
           make -j`nproc` RUST_ENABLE=y COMPILE_S_USER=32 check
 
+  systest:
+    runs-on: ubuntu-24.04-arm
+    container: ghcr.io/ivila/teaclave-trustzone-sdk-ci-runner:ubuntu-24.04
```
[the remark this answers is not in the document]
ivila: > Seems you're using a new image, does our Dockerfile need to change?

No, it uses the same Dockerfile, but this one is built for `linux/arm64`.
It seems that `yuanz0/teaclave-trustzone-sdk:ubuntu-24.04` only built for platform `linux/amd64`.
image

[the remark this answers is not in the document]
DemesneGH: Got it. Could you help to add some comments on the CI tasks for clarification? Such as:
```
# Cross-compile on AMD64 and run tests in QEMU
build-and-test-examples:

# Run systest of optee_teec and build systest of optee_utee on ARM64
systest:
```
It would help us avoid confusion in the future, as we might forget the differences.

DemesneGH:
After the tiny comments resolved please feel free to add:

DemesneGH:
BTW could the ARM64 image also help on https://github.com/apache/incubator-teaclave-trustzone-sdk/pull/160?

optee-teec/src/context.rs
```diff
@@ -57,11 +57,7 @@ impl Context {
     /// ```
     pub fn new() -> Result<Context> {
         // define an empty TEEC_Context
-        let mut raw_ctx = raw::TEEC_Context {
-            fd: 0,
-            reg_mem: false,
-            memref_null: false,
-        };
+        let mut raw_ctx = unsafe { std::mem::zeroed() };
```
DemesneGH, line 60: Since usage of `unsafe()` would need a explanation when we enable `cargo clippy` in the future. I think we need to unify the unsafe usage.
Will all similar initializations use `unsafe { std::mem::zeroed() }` or just fill in default values?
What's your opinion on it?

ivila: I prefer using `unsafe { std::mem::zeroed() }` here. While Rust marks this as unsafe because zero isn't a valid bit pattern for all types, we're only applying it to C structs where a zeroed value is always valid (consistent with how we use `memset` in C).

This also avoids unnecessary changes to the `imp` field. While this field (in `TEEC_OPERATION`, `TEEC_Session`, etc) may change over time, we explicitly shouldn't—and don't need to—care about their internal representation.

DemesneGH: Let's have a SAFETY mark above this line, the format is:
        // SAFETY:
        // explaination:xxxx
        let mut raw_ctx = unsafe { std::mem::zeroed() };

ivila: Add for `TEEC_Context` and `TEEC_Session`.

DemesneGH: LGTM

PR #180, opened 2025-04-15
ci: add build_runner_image.yml

ivila:
Support building runner images for both arm64 and amd64 via github ci.

This PR will allow developers to build a runner image through github actions.
image

After the pipeline success, the runner image can be found under the developers' packages.
image

Currently we build container with two platforms: `linux/amd64`, `linux/arm64`.
image

### DIFF
--- a/.github/workflows/build_runner_image.yml
+++ b/.github/workflows/build_runner_image.yml
@@ -1,0 +1,113 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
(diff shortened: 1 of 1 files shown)


ivila:
Please noticed that the packages would also appear under the repository.
image

So maybe we should change it to another name (maybe `teaclave-trustzone-sdk-ci-runner`?)

DemesneGH:
Thanks for this PR! Could we add some explanation about why and when developers need to build their own runner image?

ivila:
How about adding comments in the `build_runner_image.yml`:

```yml
# This action provides a standardized and reproducible way to build a CI runner
# image.
#
# It is especially helpful for developers who want to contribute to the github
# workflow of teaclave, includes the following:
#
# 1. Changes to the CI runner image via Dockerfile modifications
# 2. Experiments with new platforms (e.g., adding ARM64 runner)
#
# Developers can build their own image, test it in their fork or repository, and
# once verified, request maintainers to build and publish the official image.
# The developer’s tested image and the final official image should be identical
# to ensure consistency.
#
#
# Example Workflow:
# Suppose a developer named Alice wants to modify the base runner image.
# She would:
#
# 1. Update the Dockerfile with her changes.
# 2. Build a custom runner image using GitHub Actions.
# 3. In her branch, temporarily replace the official runner image with her
#    custom image in the CI workflow.
# 4. Run the CI using the modified setup to verify everything works as expected.
#
# Once the tests pass:
# 5. Alice opens a PR demonstrating the container used, the CI results.
# 6. The Teaclave maintainers will build an official image via GitHub Actions,
#    which should be identical to Alice's custom image.
# 7. Finally, she updates the CI workflow to use the official image built from
#    the updated Dockerfile instead of her custom one.
```

DemesneGH:
LGTM.

Issue #181, opened 2025-04-18
[TODO] Tracking on the changes of default key of OP-TEE/optee_os


ivila:
In this [#PR](https://github.com/OP-TEE/optee_os/pull/7369), OP-TEE OS updated its default key, which will cause TA verification failures in our CI. To resolve this, we need to update our QEMU images after it got merged.

jbech-linaro:
@ivila , keep an eye on [this](https://github.com/OP-TEE/optee_test/pull/792) as well. The PR's in optee_os and optee_test are dependent on each other. It's likely that you need that change as well and/or make similar changes elsewhere.

ivila:
The [PR](https://github.com/OP-TEE/optee_os/pull/7369) has been merged. After OP-TEE releases a new version(should be 4.7.0), we should update the following line:
https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/63791367dc150194609e958fc40da601e595456d/build_optee_libraries.sh#L22
to reflect the new OP-TEE version, and then rebuild the prebuilt QEMU images accordingly (the image should be given a new name to avoid affecting existing works).

PR #182, opened 2025-04-18
Add test qemu builder

ivila:
This PR adds a standardized and reproducible way to build QEMU images for our ci tests, it builds Qemu images with the following pairs:

1. (amd64, default_config)
2. (amd64, expand_memory)
3. (arm64, default_config)
4. (arm64, expand_memory)

And upload them to the artifact of the workflow, you can check an example run [here](https://github.com/ivila/incubator-teaclave-trustzone-sdk/actions/runs/14528968263):
image

This is particularly valuable for maintainers to regenerate images and keep them synchronized with OP-TEE changes, check #181 for some details.


ivila:
Mark, this is still a draft for now. I'm currently investigating a performance issue—the built result runs nearly 10 times slower than the version available at [nightlies.apache.org](https://nightlies.apache.org/teaclave/teaclave-trustzone-sdk/), which will result in failures of our tests(sleep for 5 seconds is not enough now).

ivila:
Mark: this PR still draft, there are some problems with performance and expand TA memory.
Open for early review.

ivila:
Temporarily close it to avoid needless CI running.

PR #187, opened 2025-05-09
Remove unwrap in test case

simonsso:
Compare wrapped results instead of using `unwrap()`

### DIFF
--- a/optee-teec/src/uuid.rs
+++ b/optee-teec/src/uuid.rs
@@ -149,9 +149,9 @@ mod tests {
             "11173366-2aca-19bc-beb7-10c975e6131e", // random uuid
         ];
         for origin in uuids.iter() {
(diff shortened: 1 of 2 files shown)


ivila:
Would you mind updating the tests under optee-utee as well? 
https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/0703f727cec37f167d1a3b26febb276284b07546/optee-utee/src/uuid.rs#L119-L138

Also, please add a `Signed-off-by` tag in the commit messages. 

With those minor changes, feel free to add: `Acked-by: Zehui Chen <ivila@apache.org>`

simonsso:
@ivila like this?

ivila:
> @ivila like this?

Yes, thanks.

ivila:
Some packages drop support of rustc 1.80, result in failures in STD tests. Proceed anyway.

PR #190, opened 2025-05-20
optee-utee: support TEE_GetProperty APIs

DemesneGH:
This PR introduces support for accessing TEE properties, as specified in the GlobalPlatform TEE Internal Core API Specification v1.3.1 (https://globalplatform.org/wp-content/uploads/2021/03/GPD_TEE_Internal_Core_API_Specification_v1.3.1_PublicRelease_CC.pdf):

- [4.4] Property Access Functions
- [4.5] Trusted Application Configuration Properties
- [4.6] Client Properties
- [4.7] Implementation Properties

Each property defined in Sections 4.5–4.7 of the spec has a corresponding definition in `property.rs`.

These properties enable a variety of use cases. For example, they enable retrieving metadata such as the TA description, allow a TA to determine the caller's origin (whether it was invoked by a CA or another TA), and retrieve the caller's UUID.

Please refer to the `property-rs` example for a demonstration of these use cases.


optee-utee/src/property.rs
```diff
@@ -0,0 +1,481 @@
...
+// 
+// To get the property value, use the `get` method.
+// Example usage:
+// 
+// ``` no_run
+// use optee_utee::{PropertyKey, CurrentTaAppId};
+// 
+// let my_property = CurrentTaAppId.get()?;
+// ```
+define_property_key!(CurrentTaAppId, CurrentTa, "gpd.ta.appID", Uuid);
```
m4sterchain, line 290: The idea here is to abstract the complexity of fetching properties under Property Sets by introducing dedicated types. The main concern is the naming convention used for these newly introduced types.

The suggested approach is to map type names directly from the corresponding property keys (or their suffixes). For example:

gpd.ta.doesNotCloseHandleOnCorruptObject → TaDoesNotCloseHandleOnCorruptObject

gpd.tee.apiversion → TeeApiVersion

gpd.ta.dataSize → TaDataSize

This naming convention allows developers to intuitively derive the type name from the GlobalPlatform documentation without needing to inspect the implementation details. It significantly improves developer experience and code readability.

examples/property-rs/ta/src/main.rs
```diff
@@ -0,0 +1,91 @@
...
+    // check caller
+    let client_identity = CurrentClientIdentity.get()?;
+    trace_println!(
+        "[+] TA get caller identity: login: {}, uuid: {}",
+        client_identity.login_type(),
+        client_identity.uuid()
+    );
+
+    // test the other property:
+    let core_version = TeeImplementationInternalCoreVersion.get()?;
```
m4sterchain, line 63: - As this logic serves as test case. Can we add assertion to the return value?

examples/property-rs/ta/build.rs
```diff
@@ -0,0 +1,26 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+use optee_utee_build::{Error, RustEdition, TaConfig};
+use proto;
+
+fn main() -> Result<(), Error> {
+    let config = TaConfig::new_default_with_cargo_env(proto::UUID)?
+        .ta_data_size(1 * 1024 * 1024)
+        .ta_stack_size(1 * 1024 * 1024);
```
ivila, line 24: Can we remove these? Default settings should be OK?

optee-utee/src/uuid.rs
```diff
@@ -79,6 +79,11 @@ impl Uuid {
         ))
     }
 
+    /// Creates a `Uuid` from a raw `TEE_UUID` object.
+    pub fn from_raw(raw: raw::TEE_UUID) -> Uuid {
```
ivila, line 83: Can we use a From trait instead?

optee-utee/src/uuid.rs
```diff
@@ -79,7 +79,12 @@ impl Uuid {
         ))
     }
 
-    /// Crates a raw TEE client uuid object with specified parameters.
+    /// Creates a `Uuid` from a raw `TEE_UUID` object.
+    // pub fn from_raw(raw: raw::TEE_UUID) -> Uuid {
+    //     Self { raw }
+    // }
```
DemesneGH, line 85: All above comments are resolved. BTW forgot to delete these lines, will delete in next commit for adding review tags.

optee-utee/src/identity.rs
```diff
@@ -0,0 +1,76 @@
...
+    Public = raw::TEE_LOGIN_PUBLIC,
+    User = raw::TEE_LOGIN_USER,
+    Group = raw::TEE_LOGIN_GROUP,
+    Application = raw::TEE_LOGIN_APPLICATION,
+    ApplicationUser = raw::TEE_LOGIN_APPLICATION_USER,
+    ApplicationGroup = raw::TEE_LOGIN_APPLICATION_GROUP,
+    TrustedApp = raw::TEE_LOGIN_TRUSTED_APP,
+}
+
+impl fmt::Display for LoginType {
```
ivila, line 64: Maybe just use `strum` and derive [strum::Display](https://docs.rs/strum/latest/strum/derive.Display.html)

optee-utee/src/property.rs
```diff
@@ -0,0 +1,483 @@
...
+                        &mut out_size,
+                    )
+                };
+                if res != raw::TEE_SUCCESS {
+                    return Err(Error::from_raw_error(res));
+                }
+
+                // For C raw strings, the last byte is a null terminator
+                // So we need to trim it
+                Ok(String::from_utf8_lossy(&out_buffer).trim_end_matches('\0').to_string())
```
ivila, line 95: I would suggest using `core::ffi::CStr` instead of handling the null terminator ourself.
Maybe [CStr::from_bytes_with_nul](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_bytes_with_nul)?

ivila:
`Reviewed-by: Zehui Chen <ivila@apache.org>` 😀

PR #192, opened 2025-05-22
crates: move secure_db to shared crates/ directory

DemesneGH:
This PR reorganizes basic modules to promote reuse and improve the development experience for TAs.

* In this initial stage, the `secure_db` module has been moved into the `crates` directory.
* It is now shared across both the `secure_db_abstraction` and `eth_wallet` examples.

Future Plans:

* More common modules can be moved into `crates` to encourage code reuse and simplify TA development.
* Once the structure stabilizes, these crates are planned to be published on `crates.io`, making them easily accessible to the community.


ivila:
`Acked-by: Zehui Chen <ivila@apache.org>` 😀

projects/web3/eth_wallet/ta/src/main.rs
```diff
@@ -18,37 +18,53 @@
...
+pub struct WalletSession {
+    db_client: SecureStorageClient,
+}
+
+impl Default for WalletSession {
+    fn default() -> Self {
+        Self {
+            db_client: SecureStorageClient::open(DB_NAME)
+                .expect("Failed to create SecureStorageClient"),
+        }
```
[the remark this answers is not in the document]
m4sterchain: Please fix the comment from Copilot

projects/web3/eth_wallet/ta/src/main.rs
```diff
@@ -18,37 +18,53 @@
...
+
 #[ta_create]
 fn create() -> optee_utee::Result<()> {
     trace_println!("[+] TA create");
     Ok(())
 }
 
 #[ta_open_session]
-fn open_session(_params: &mut Parameters) -> optee_utee::Result<()> {
+fn open_session(_params: &mut Parameters, _sess_ctx: &mut WalletSession) -> optee_utee::Result<()> {
```
m4sterchain, line 57: What is the purpose of this unused parameter?

projects/web3/eth_wallet/ta/src/main.rs
```diff
@@ -83,21 +101,27 @@ fn create_wallet(_input: &proto::CreateWalletInput) -> Result<proto::CreateWalle
...
 }
 
-fn derive_address(input: &proto::DeriveAddressInput) -> Result<proto::DeriveAddressOutput> {
-    let secure_object = load_from_secure_storage(input.wallet_id.as_bytes())
+fn derive_address(
+    db_client: &SecureStorageClient,
+    input: &proto::DeriveAddressInput,
+) -> Result<proto::DeriveAddressOutput> {
+    let wallet = db_client
+        .get::<Wallet>(&input.wallet_id.to_string())
```
m4sterchain, line 121: &input.wallet_id

projects/web3/eth_wallet/ta/src/main.rs
DemesneGH: Add explanation as code comment:
// The _sess_ctx: &mut WalletSession is explicitly defined here to fit into the
// macros definition (optee-utee/macros/src/lib.rs), meaning that we have a
// context for the session which would be initialized in the open_session() method.
// The context is initialized automatically by calling WalletSession::default()
// in the macro.

projects/web3/eth_wallet/ta/src/main.rs
DemesneGH:
```
    pub fn get<V>(&self, key: &V::Key) -> Result<V>
    where
        V: Storable + serde::de::DeserializeOwned,
        V::Key: Into<String> + Clone,
```
The StorageKey type (`V::Key`) should impl `Into<String>`, since the Uuid is from the third-party crate `uuid`, we cannot impl trait for it.
Here I convert the Uuid into String and use the String as StorageKey type. But we need to `&input.wallet_id.to_string()`. To eliminate this maybe we can define a wrapper struct `UuidKey(Uuid)` which is also not elegant, or our abstraction needs to be improved for this case.

DemesneGH: Fixed in the latest commit: use ToString instead of Into<String>.

projects/web3/eth_wallet/ta/src/main.rs
DemesneGH: Given that the pattern also exists in other examples: https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/main/examples/acipher-rs/ta/src/main.rs#L32-L54, also in `hotp-rs` `aes-rs` `digest-rs` and so on, I think it's reasonable to use the same pattern here for now, we can open another PR to revise the session context handling then update all related examples.
If it's not acceptable, I'm okay to rollback the `eth_wallet` example to the version doesn't use the secure_db crate.

@m4sterchain Which approach would you prefer?

m4sterchain: No problem. We can proceed and resolve `sessionContext` together in future PRs.

projects/web3/eth_wallet/ta/src/main.rs
```diff
@@ -18,37 +18,66 @@
...
+            .map_err(|e| anyhow!("Failed to create SecureStorageClient: {:?}", e))?;
+        Ok(Self { db_client })
+    }
+}
+
+impl Default for WalletSession {
+    fn default() -> Self {
+        Self::new().unwrap_or_else(|e| {
+            trace_println!("Error initializing WalletSession: {:?}", e);
+            panic!("Failed to initialize WalletSession");
```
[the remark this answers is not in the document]
m4sterchain: @DemesneGH could you please fix the copilot comments, including useful ones in  `Comments suppressed due to low confidence (3)`? Thanks.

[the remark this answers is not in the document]
DemesneGH: Since we cannot eliminate the panic if we use current session context API, `eth_wallet` is updated to not use the shared session context.

m4sterchain:
> Any other comments before merging? @m4sterchain 

Feel free to merge after all comments are resolved. Thanks.

PR #195, opened 2025-06-02
Expose login field from raw::TEEC_OpenSession

simonsso:
Adding a new method to open a session with login field set by caller instead of hard coding it to default (`ConnectionMethods::LoginPublic`)

Legacy constructor is not removed.

### DIFF
--- a/optee-teec/src/context.rs
+++ b/optee-teec/src/context.rs
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use crate::{raw, Error, Operation, Param, ParamNone, Result, Session, Uuid};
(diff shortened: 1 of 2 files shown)


DemesneGH:
LGTM.
@ivila Any comments?

simonsso:
Could we merge and release the crate or do we need something more?

ivila:
> Could we merge and release the crate or do we need something more?

The changes have been merged, but we may not publish a new crate version immediately. Please refer to the release schedule here: https://github.com/apache/incubator-teaclave-trustzone-sdk/issues/165.

As a temporary workaround, you can override the optee_teec crate with the GitHub repository in your Cargo.toml: https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html?highlight=patch#the-patch-section

PR #197, opened 2025-06-06
Refine license of example mnist-rs

ivila:


examples/mnist-rs/ta/common/src/model.rs
```diff
@@ -55,6 +55,10 @@ impl<B: Backend> Model<B> {
 }
 
 impl<B: Backend> Model<B> {
+
+    // Originally inspired by the burn/examples/mnist-inference-web package. You
+    // may refer to https://github.com/tracel-ai/burn/blob/v0.16.0/examples/mnist-inference-web/src/web.rs
+    // for details.
```
DemesneGH, line 61: Seems we also need the similar notice in `trainer.rs`?

ivila: The notice has been present in trainer.rs since the first commit. I'm adding this line now because I realized I missed including the notice in #173.

Here are the notice in `trainer.rs`: 
https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/main/examples/mnist-rs/ta/train/src/trainer.rs#L51-L114

ivila:
The CI `[test-on-arm64-host / build-and-test-examples-for-32bit-std-TAs](https://github.com/apache/incubator-teaclave-trustzone-sdk/actions/runs/15482211571/job/43589998905?pr=197#logs)` failed by timeout in fetch crates.

DemesneGH:
Merged, thanks!

PR #209, opened 2025-07-10
Refactor Object

ivila:
There are 5 commits in this pull request:
1) commits to split files: just want to keep the codes clean, make some changes on `ObjectHandle`(make its function `pub`) to fix compilation errors.
2) introduce `optee-utee-mock` crate: I think we really need a mocking crate for unittests.
3) refactor object: **the main part of this PR**.
4) fix compilation errors in `crates` and `examples`: see the breaking changes.
5) fix documents

### DIFF
--- a/crates/secure_db/Cargo.toml
+++ b/crates/secure_db/Cargo.toml
@@ -27,3 +27,4 @@ optee-utee = { path = "../../optee-utee" }
 bincode = "1.3.3"
 anyhow = "1.0" 
 serde = { version = "1.0", features = ["derive"] }
+hashbrown = { version = "0.15.4", features = ["serde"] }
--- a/crates/secure_db/src/db.rs
+++ b/crates/secure_db/src/db.rs
@@ -17,7 +17,8 @@
 
 use crate::{delete_from_secure_storage, load_from_secure_storage, save_in_secure_storage};
 use anyhow::{bail, ensure, Result};
-use std::collections::{HashMap, HashSet};
+use hashbrown::HashSet;
+use std::collections::HashMap;
 
 // SecureStorageDb is a key-value storage for TA to easily store and retrieve data.
 // First we store the key list in the secure storage, named as db_name.
@@ -40,6 +41,10 @@ impl SecureStorageDb {
                 // create new db
                 Ok(Self {
                     name,
+                    // Note: `std::collections::HashSet` was replaced with
+                    // `hashbrown::HashSet`, due to a write permission fault
+                    // observed during testing. The exact cause of the issue is
+                    // unclear, but using `hashbrown::HashSet` resolves it.
                     key_list: HashSet::new(),
                 })
             }


ivila:
> Hi @ivila,
> 
> Thanks for the implementation! I’ve reviewed the `ObjectHandle` abstraction and would like to confirm my understanding — please correct me if anything is off:
> 
> * `raw::TEE_ObjectHandle` is the low-level FFI type used at the Rust-C boundary.
> * `ObjectHandle` is the high-level Rust wrapper that encapsulates the FFI handle, offering a safer and more idiomatic interface.
> * The conversion currently always succeed: `ObjectHandle` can be constructed from `raw::TEE_ObjectHandle`, and raw access is only available via `handle()` or `handle_mut()`.
> 
> Given this design, I’d like to suggest revisiting the decision to support a "null" handle (`ObjectHandle::new_null()`) in safe Rust code.
> 
> The main use case, as I understand, is that a handle may initially be `null` when obtained from FFI (e.g., via out-params in C APIs). That makes sense at the raw layer. However, `ObjectHandle` is a Rust abstraction, and ideally, it should enforce the invariant that **only valid (non-null) handles** are held inside. This aligns with Rust’s convention: if a value exists, it should be valid by construction.
> 
> ### Suggestion:
> To better leverage Rust’s type system, could we refine the design as follows?
> 
> * **Remove** the `ObjectHandle::new_null()` constructor.
> * Introduce `TryFrom<raw::TEE_ObjectHandle>` implementation for `ObjectHandle`, where we check for null and return a `Result<Self, Error>`.
> * Ensure that all entry points from the FFI boundary perform this validation before returning an `ObjectHandle`.
> 
> This would make `ObjectHandle` strictly non-null and eliminate the need for an `is_null()` method entirely in most cases. It also communicates clearly to developers that once they have an `ObjectHandle`, it’s safe to use without extra checks.
> 
> The code pattern might be like this after the refinement.
> 
> ```
> let mut raw_handle: raw::TEE_ObjectHandle = core::ptr::null_mut();
> 
> let status = unsafe {
>     raw::TEE_OpenPersistentObject(
>         storage_id as u32,
>         object_id.as_ptr() as _,
>         object_id.len(),
>         flags.bits(),
>         &mut raw_handle,
>     )
> };
> 
> if status == raw::TEE_SUCCESS {
>     // Convert to safe Rust wrapper
>     return ObjectHandle::try_from(raw_handle);
> } else {
>     return Err(Error::from_raw_error(status));
> }
> ```
> 
> Of course, I’m not entirely sure whether this would cover **all** use cases — feel free to share your thoughts or if there are edge cases we should consider.

Agreed. The same issues are occurring with TransientObject as well—and since TransientObject relies heavily on ObjectHandle::new_null(), I’ll investigate further and determine what changes are needed for TransientObject.

optee-utee/src/object/persistent_object.rs
```diff
@@ -529,26 +484,186 @@ impl PersistentObject {
...
+                    move |_, _, _, _, _, _, _, obj: *mut raw::TEE_ObjectHandle| {
+                        unsafe {
+                            *obj = *handle;
+                        }
+                        raw::TEE_SUCCESS
+                    },
+                )
+                .once();
+        }
+
```
DemesneGH: I agree we need a `optee-utee-mock` crate to mock the raw C APIs then we can test the modules without the  real OP-TEE OS. Seems each testcase follows similar pattern and there're duplicated code which makes it a few complicated, any way to define the shared mock.expect and reuse in each testcase? Something like:
```
expect_create_success(..)
expect_close_success(..)
```
Then in testcases we only focus on the specific behavior or logic under test.

optee-utee/src/object/object_handle.rs
```diff
@@ -0,0 +1,185 @@
...
+        assert!(obj.is_null());
+    }
+
+    /// Ensures the return value of `handle_mut` can be written.
+    ///
+    /// TODO: still need a systest to ensure that, I found that if change
+    /// `ObjectHandle` from `Object(Box<raw::TEE_ObjectHandle>)` to
+    /// `ObjectHandle(raw::TEE_ObjectHandle)`, this unittest still works,
+    /// but will result in `Panic by Permission Fault` in TA, so things is kind
+    /// of different between unittest and TA.
```
DemesneGH: I'm not clear on this issue yet. Maybe we can revisit it after trying the revising about `null()`.

ivila: It means that some memory issues are un-detected in unit tests.

If `ObjectHandle` use the following definition:
```rust
struct ObjectHandle(raw::TEE_ObjectHandle);
```
presents a memory issue because the inner `raw::TEE_ObjectHandle` resides in protected memory (stack/TEE-OS space), writing to it is allowed in unit tests but forbidden in TEE-OS environments.

So calling handle_mut() to get a mutable pointer and write through it:

* Works in unit tests (false sense of safety)
* Causes panics in TEE-OS (memory protection fault)

optee-utee/src/object/persistent_object.rs
ivila: I think the complexity arises from Rust's syntax, and the logic cannot be defined in a straightforward way due to its context-awareness. For instance, in this case, we need to generate a unique `raw::TEE_ObjectHandle`. When calling `TEE_CreatePersistentObject`, we provide this value. Later, when calling `TEE_CloseObject`, we must ensure the correct value is passed—not just any random value.

optee-utee/src/object/object_handle.rs
DemesneGH: Thanks for the clarification, the memory issue makes sense to me. But why we need always expose the `handle_mut()`?
From what I understand, the typical workflow is: we receive a handle through open or create, use it to interact with other raw APIs, and finally close it. Once we receive the handle from raw, we will not change the value of it on it's lifetime. If so, how about:
```
// create a temporary variable to receive the handle
let mut handle: raw::TEE_ObjectHandle = core::ptr::null_mut();
let result = unsafe {
    raw::TEE_OpenPersistentObject(
        storage_id as u32,
        object_id.as_ptr() as _,
        object_id.len(),
        flags.bits(),
        &mut handle,
    )
};
match result {
    // construct the handle, e.g. from_raw(), or .try_into() or other fn you'd like
    raw::TEE_SUCCESS => Ok(ObjectHandle::from_raw(handle)),
    code => Err(Error::from_raw_error(code)),
}
```
And no need to expose the handle_mut().

ivila: Actually, the `new_null` and `handle_mut` functions are combined for internal use (i.e., within `TEE_OpenPersistentObject` in PersistentObject). I’m using this approach to avoid code duplication and provide standardize usage, which helps prevent memory-related bugs.

For example, the demo code you’ve provided demonstrates the memory issue I mentioned earlier, which will result in a **Permission Fault** in OP-TEE OS.

It should be written like this:
```rust
// create a temporary variable to receive the handle
// which must be in the Box.
let mut handle: Box<raw::TEE_ObjectHandle> = Box::new(core::ptr::null_mut());
let result = unsafe {
    raw::TEE_OpenPersistentObject(
        storage_id as u32,
        object_id.as_ptr() as _,
        object_id.len(),
        flags.bits(),
        &mut handle,
    )
};
match result {
    // construct the handle, e.g. from_raw(), or .try_into() or other fn you'd like
    // how should we pass the inner handle? Should we forget the box to prevent Drop?
    raw::TEE_SUCCESS => Ok(ObjectHandle::from_raw(handle)),
    code => Err(Error::from_raw_error(code)),
}
```

Maybe I should make both the new_null and handle_mut functions pub(crate) instead. This way:

* We can ensure that only valid ObjectHandle values are provided to developers.
* We won’t need to make significant changes to TransientObject (which should be done eventually, but isn’t the focus of this PR as it heavily related to `crypto_op`).

optee-utee/src/object/persistent_object.rs
DemesneGH: How about defining macros:
```
macro_rules! expect_create_success {
    ($mock:expr, $handle:expr) => {{
        let h = $handle.clone();
        $mock.expect_TEE_CreatePersistentObject()
            .return_once_st(move |_, _, _, _, _, _, _, obj| {
                unsafe {
                    *obj = *h.get();
                }
                raw::TEE_SUCCESS
            })
            .once();
    }};
}
macro_rules! expect_create_fail {
    ($mock:expr, $code:expr) => {{
        $mock.expect_TEE_CreatePersistentObject()
            .return_once_st(move |_, _, _, _, _, _, _, _| $code)
            .once();
    }};
}
...
```
In testcases:
```
#[test]
fn test_create_and_successfully_close_delete() {
    let _lock = SERIAL_TEST_LOCK.lock();

    let mut mock = MockObjectController::new();
    let handle = std::sync::Arc::new(std::cell::UnsafeCell::new(core::ptr::null_mut()));

    expect_create_success!(mock, handle);
    expect_close_and_delete_success!(mock, handle);

    set_global_object_mock(mock);

    let obj = PersistentObject::create(
        ObjectStorageConstants::Private,
        &[],
        DataFlag::ACCESS_WRITE,
        None,
        &[],
    )
    .expect("it should be ok");

    obj.close_and_delete().expect("it should be ok");
}
```
If it makes sense, how about defining the macros in `optee-utee-mock` crate? This would keep test cases cleaner, and ensure the basic mock expectations remain with the mock implementation.

optee-utee/Cargo.toml
```diff
@@ -40,6 +40,7 @@ serde = { version = "1.0.215" }
 serde_json = { version = "1.0.133" }
 # disable linking when running unit tests
 optee-utee-sys = { version = "0.5.0", path = "optee-utee-sys", features = ["no_link"] }
+optee-utee-mock = { version = "0.5.0", path = "optee-utee-mock" }
```
DemesneGH, line 43: Consider marking this crate as optional and enabling it only for tests? Seems it's not intended for normal developers or production use for now.

ivila: I don't understand, this is dev-dependencies, unrelated to developers or production use.

optee-utee/src/object/persistent_object.rs
ivila: Done by adding functions for `MockObjectController` like `new_valid_test_handle`

optee-utee/src/object/transient_object.rs
```diff
@@ -0,0 +1,453 @@
...
+    /// 5) If the Implementation detects any other error associated with this
+    ///    function which is not explicitly associated with a defined return
+    ///    code for this function.
+    pub fn generate_key(&self, key_size: usize, params: &[Attribute]) -> Result<()> {
+        let p: Vec<raw::TEE_Attribute> = params.iter().map(|p| p.raw()).collect();
+        unsafe {
+            match raw::TEE_GenerateKey(
+                self.handle(),
+                key_size as u32,
+                p.as_slice().as_ptr() as _,
```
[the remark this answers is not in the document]
ivila: Not the target of this PR, skip it.

optee-utee/src/object/object_define.rs
```diff
@@ -0,0 +1,128 @@
...
+    DataSeekSet,
+    /// The data position is set to its current position plus offset.
+    DataSeekCur,
+    /// The data position is set to the size of the object data plus offset.
+    DataSeekEnd,
+}
+
+impl Into<raw::TEE_Whence> for Whence {
+    fn into(self) -> raw::TEE_Whence {
+        match self {
```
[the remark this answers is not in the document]
ivila: skip

optee-utee/src/object/object_info.rs
```diff
@@ -0,0 +1,78 @@
...
+    ///       differ.
+    ///     * For a [TransientObject](crate::TransientObject), set to 0.
+    /// 7) `handleFlags`: A bit vector containing one or more
+    ///    [HandleFlag](crate::HandleFlag) or [DataFlag](crate::DataFlag).
+    pub fn from_raw(raw: raw::TEE_ObjectInfo) -> Self {
+        Self { raw }
+    }
+
+    /// Return the `dataSize` field of the raw structure `TEE_ObjectInfo`.
+    pub fn data_size(&self) -> usize {
```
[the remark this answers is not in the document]
ivila: skip

ivila:
@DemesneGH unit tests updated in `object_handle.rs` for demo, and left unchanged in `persistent_object.rs` and `transient_object.rs`(will changed after the logic is accepted)

optee-utee/Cargo.toml
DemesneGH: Sorry for mistake. I didn't notice it's in [dev-dependencies].

DemesneGH:
> @DemesneGH unit tests updated in `object_handle.rs` for demo, and left unchanged in `persistent_object.rs` and `transient_object.rs`(will changed after the logic is accepted)

The testcase changes look good to me, thanks!

optee-utee/src/object/object_handle.rs
DemesneGH: > For example, the demo code you’ve provided demonstrates the memory issue I mentioned earlier, which will result in a **Permission Fault** in OP-TEE OS.

Do you mean that the variable used to receive the handle cannot be on the stack (causes Permission Fault) and should be on the heap instead?
I notice that the C examples for `object` also use the variable on the stack: https://github.com/linaro-swg/optee_examples/blob/master/secure_storage/ta/secure_storage_ta.c#L85. Any thing different with Rust?

Additionally, I recall the similar pattern from the TA invoke command implementation also uses variables on stack:
https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/main/optee-utee/src/ta_session.rs#L52

ivila: > > For example, the demo code you’ve provided demonstrates the memory issue I mentioned earlier, which will result in a **Permission Fault** in OP-TEE OS.
> 
> Do you mean that the variable used to receive the handle cannot be on the stack (causes Permission Fault) and should be on the heap instead? I notice that the C examples for `object` also use the variable on the stack: https://github.com/linaro-swg/optee_examples/blob/master/secure_storage/ta/secure_storage_ta.c#L85. Any thing different with Rust?
> 
> Additionally, I recall the similar pattern from the TA invoke command implementation also uses variables on stack: https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/main/optee-utee/src/ta_session.rs#L52

I haven’t explored it fully yet, but there’s likely some distinction. You can check it here:

* Change `ObjectHandle(Box<raw::TEE_ObjectHandle>)` to `ObjectHandle(raw::TEE_ObjectHandle)`
    changes: https://github.com/ivila/incubator-teaclave-trustzone-sdk/commit/8edc7c3b36001a1226fec33ccd74db047499d80b
    pipeline failed by write permission fault: https://github.com/ivila/incubator-teaclave-trustzone-sdk/actions/runs/16263572387/job/45914118660
  
image

* Use codes like your demo
    changes: https://github.com/ivila/incubator-teaclave-trustzone-sdk/commit/7abbc6560d91ca4d33ce0ec1a0c19436c85503fc
    pipeline failed by write permission fault: https://github.com/ivila/incubator-teaclave-trustzone-sdk/actions/runs/16263548106/job/45914044273
   
image

You can test it locally using `secure_db_abstraction-rs`. If you've previously run it successfully, remember to `rm -rf /data/tee` before testing the new compiled TA again.

ivila: @DemesneGH 
After digging into it, I found something strange: the `write permission fault` is actually caused by this line:
https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/06466163ceb24813eb647aa576550c09676f7418/crates/secure_db/src/db.rs#L43

After replacing `std::collections::HashSet` with `hashbrown::HashSet`, the problem was resolved.

I suspect that when using `std::collections::HashSet`, some kind of stack overlap or memory conflict occurred.

DemesneGH: > After replacing `std::collections::HashSet` with `hashbrown::HashSet`, the problem was resolved.

Thanks so much for digging into this and identifying the root cause. Seems it's a bug in our manually ported std. I've created issue https://github.com/apache/incubator-teaclave-trustzone-sdk/issues/210 to track it and will look into it soon.

optee-utee/src/object/persistent_object.rs
```diff
@@ -0,0 +1,643 @@
...
+        let mut handle: raw::TEE_ObjectHandle = core::ptr::null_mut();
+        let handle_mut = &mut handle;
+        match unsafe {
+            raw::TEE_OpenPersistentObject(
+                storage_id as u32,
+                object_id.as_ptr() as _,
+                object_id.len(),
+                flags.bits(),
+                handle_mut,
+            )
```
DemesneGH, line 98: Consider removing line90 and line176 for simplicity?

ivila: I tend to move codes out of `unsafe` as more as possible， because in `unsafe` we lose the ability of rust checker and can easily lead to bugs like this: https://github.com/apache/incubator-teaclave-trustzone-sdk/pull/120#issuecomment-2559150989

DemesneGH: Got it. Could we have some comment above these lines to explain the motivation? Because in the future we may forget why we write like this.

optee-utee/src/object/transient_object.rs
```diff
@@ -0,0 +1,427 @@
...
+///
+/// Contrast [PersistentObject](crate::PersistentObject).
+#[derive(Debug)]
+pub struct TransientObject(ObjectHandle);
+
+impl TransientObject {
+    /// Create an object with a null handle which points to nothing.
+    pub fn null_object() -> Self {
+        Self(ObjectHandle::new_null())
+    }
```
DemesneGH: Looks like `null_object()` is used in the crypto examples to initialize the session context, e.g.https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/main/examples/acipher-rs/ta/src/main.rs#L39. Since refactoring the session context is planned but not urgent, I agree we keep `null_object()` for now and leave it for a future PR.

@ivila Could you help to add a brief comment above `null_object()` explaining its current purpose and usage context?

optee-utee/src/object/persistent_object.rs
```diff
@@ -0,0 +1,643 @@
...
+        object_id: &[u8],
+        flags: DataFlag,
+        attributes: Option<ObjectHandle>,
+        initial_data: &[u8],
+    ) -> Result<Self> {
+        let mut handle: raw::TEE_ObjectHandle = core::ptr::null_mut();
+        let handle_mut = &mut handle;
+        let attributes = match attributes {
+            Some(a) => a.handle(),
+            None => ptr::null_mut(),
```
DemesneGH: Seems there're multiple source of `null_mut()`: `core::ptr::null_mut()`, `ptr::null_mut()`(use core::ptr) and `std::ptr::null_mut()`. Could we unify this?

DemesneGH:
LGTM, thanks!

Issue #174, opened 2025-03-18
[TODO] Update for CI checking script in OP-TEE


DemesneGH:
The CI in the OP-TEE repo runs an expect script to test Rust examples. Originally, this script was located at `teaclave-trustzone-sdk/ci/qemu-check.exp` but has been duplicated to `OP-TEE/build/rust.exp` in https://github.com/OP-TEE/build/pull/717 and maintained in OP-TEE repo.

However, `rust.exp` needs to be updated to include checks for additional `no_std` examples: `tcp_client` and `udp_client`.  

Before adding these checks, we should first enable IPv6 support in QEMU, as discussed 
 in https://github.com/apache/incubator-teaclave-trustzone-sdk/pull/164#discussion_r1928165412, to properly test the `tcp_client` example.  

Leaving a TODO here as a record.

ivila:
@DemesneGH 
May I know what changes you made to `qemu-system-aarch64` in `optee-qemuv8-4.2.0-ubuntu-24.04-expand-ta-memory`? 
I'm currently trying to add the ARM64 platform to the CI, but I'm running into some issues with it. (The none `expand-ta-memory` variant works fine so far.)

![Image](https://github.com/user-attachments/assets/21f36798-66b3-4651-b9c3-db641c6d0c9c)

PS: part of the error log
```shell
D/TC:0   add_phys_mem:666 VCORE_UNPG_RX_PA type TEE_RAM_RX 0x0e100000 size 0x00085000
D/TC:0   add_phys_mem:666 VCORE_UNPG_RW_PA type TEE_RAM_RW 0x0e185000 size 0x0017b000
D/TC:0   add_phys_mem:666 ta_base type TA_RAM 0x0e300000 size 0x01d00000
D/TC:0   add_phys_mem:666 GIC_REDIST_BASE type IO_SEC 0x08000000 size 0x01000000
D/TC:0   add_phys_mem:666 GICD_BASE type IO_SEC 0x08000000 size 0x00200000
D/TC:0   add_phys_mem:680 Physical mem map overlaps 0x8000000
D/TC:0   add_phys_mem:666 GICC_BASE type IO_SEC 0x08000000 size 0x00200000
D/TC:0   add_phys_mem:680 Physical mem map overlaps 0x8000000
D/TC:0   add_phys_mem:666 CONSOLE_UART_BASE type IO_SEC 0x09000000 size 0x00200000
[...]

PR #211, opened 2025-07-15
update optee-version

### LINKED ISSUE #181: [TODO] Tracking on the changes of default key of OP-TEE/optee_os

ivila:
* ci: remove needless ipv6 patch
* scripts: use optee 4.7.0

closes #181


ivila:
CI failed by network issue:
image

m4sterchain:
The `OPTEE-repo-build-and-run-examples-*-TAs` workflow tests the latest Teaclave TrustZone SDK commits against the [OP-TEE master branch from scratch](https://github.com/OP-TEE/manifest/blob/master/common.xml). The workflow consumes resources and facing network instability issues.

However, I believe this approach is unnecessary. The Rust SDK should prioritize compatibility with **stable OP-TEE OS releases**, as testing and maintaining compatibility with the OP-TEE OS development branch should not be the primary focus of the Teaclave TrustZone SDK.

Focusing on stable releases will significantly streamline the daily development workflow. For example:
- A stable OP-TEE release triggers a QEMU image build.
- A QEMU image change prompts the republishing of the development Docker image, pinned to a specific OP-TEE version, supporting TA development and CI testing in both standard and no-std environments.
- Teaclave SDK development and CI rely on this stable development Docker image.

This approach mitigates the instability of the current CI workflow by leveraging a consistent, version-pinned Docker image.
As the core maintainer of the CI pipelines @ivila @DemesneGH, I'd like to hear your feedback on this and align our focus of maintenance.

ivila:
Merge anyway.

DemesneGH:
> It’s good to keep the Docker image available, but as a quick solution, perhaps we should either `remove these specific jobs` or `only trigger them when changes are merged into main`.

`only trigger them when changes are merged into main` looks good to me for now. 

As discussed in https://github.com/apache/incubator-teaclave-trustzone-sdk/issues/174, the test script has not been updated for a while, meaning we cannot test the new examples, limiting the value of this test job.
However, the value of this CI job, is to test the if our SDK compatible with the latest OP-TEE, which could help to catch the issue early. If it doesn't block our contribution workflow, it’s better to have this check than not. Reducing the frequency of trigger this job makes sense for me.

> * A stable OP-TEE release triggers a QEMU image build.
> * A QEMU image change prompts the republishing of the development Docker image, pinned to a specific OP-TEE version, supporting TA development and CI testing in both standard and no-std environments.
> * Teaclave SDK development and CI rely on this stable development Docker image.

It’s great to have a clear and streamlined daily workflow, but keeping the OP-TEE CI job isn’t a bad thing either.

PR #212, opened 2025-07-16
ci: separate tests

### LINKED ISSUE #191: [RFC] Adjust CI Strategy for Testing in OP-TEE Repository
Currently, we maintain two types of CI pipelines (on multiple ARCH):

1. **Test in SDK**:

   * Builds Rust examples and runs test scripts under `sdk/tests/`
   * Executes tests using a prebuilt OP-TEE image (pinned to a specific release) and QEMU

2. **Test in OP-TEE Repo**:

   * The SDK exists as a submodule inside the OP-TEE repository
   * CI fetches the latest OP-TEE source, overwrites `optee_repo/optee_rust` with the latest SDK code, and runs OP-TEE-maintained checks (some `expect` scripts, also discussed in https://github.com/apache/incubator-teaclave-trustzone-sdk/issues/174)

### Discussion Topic

This proposal focuses on **"Test in OP-TEE Repo"** pipeline.

**Pros:**

* Ensures our latest SDK changes remain compatible with the latest OP-TEE repository code
* Helps catch and address integration issues early, reduces our effort when we need to pin a new SDK version in OP-TEE manifest

**Cons:**

* Building the full OP-TEE images from source significantly increases CI time, which is inefficient given our frequent updates
* If the `expect` scripts in the OP-TEE repo aren’t updated, only old examples are tested, reducing the value of the CI run (https://github.com/apache/incubator-teaclave-trustzone-sdk/issues/174)
* The OP-TEE repo already pins a specific SDK version via the https://github.com/OP-TEE/manifest/blob/master/qemu_v8.xml, so testing integration on every SDK commit may be unnecessary

Based on the above considerations, how about revise the workflow as follows:

* **On each SDK change (e.g. pull request):**

  * Only run the **Test in SDK** pipelines to validate functionality against the pinned OP-TEE release
  * Disable the **Test in OP-TEE Repo** for each new PR

* **When preparing an SDK release: (or we wants to pin the new SDK version in OP-TEE manifest)**

  * Run the **Test in OP-TEE Repo** pipeline once to ensure compatibility with the latest OP-TEE code (it can be one of the steps of a CI specific for checking/packing release), update `expect` scripts locally for testing (if needed)
  * Submit a PR to update the OP-TEE manifest to pin the new SDK version
  * Submit a PR to update the `expect` scripts in the OP-TEE repository (if needed)

It helps shorten our CI duration while clarifying the integration process with OP-TEE.

ivila:
closes #191 

Test in OP-TEE Repo only when merged into main.


ivila:
Maybe we should stop running it automatically and instead move it into a dedicated release pipeline, such as a Release Check stage.

DemesneGH:

> Maybe we should stop running it automatically and instead move it into a dedicated release pipeline, such as a Release Check stage.

I think we can merge this PR and adjust it to the `Release Check stage` in future follow-up PRs.

PR #27, opened 2021-05-21
Add support for tee-supplicant loadable plugins

DemesneGH:
Hi @mssun 
The loadable plugins framework is a new feature in OPTEE 3.13.0. It allows TA actively invoke a command defined in normal world plugins (which are compiled as .so dynlibs).See [this commit](https://github.com/OP-TEE/optee_client/commit/1e91cc797f72261fb03e7c2ce30944ff613ac07e).
In this commit I've updated:
1. In`optee-teec` `optee-utee` and `optee-utee-sys`:Wrap for related apis;
2. Add a new example `supp_plugin` for tee-supplicant plugins.The work logic is `host->TA->plugin`.Plugins are identified by uuid.

Best Regards,
Yuan

PR #221, opened 2025-08-15
examples: tls client and server for new rustls

DemesneGH:
This PR mainly updates the TLS client and TLS server examples to support Rustls **0.23.0+** (currently using `0.23.12`).  

Upgrading Rustls and its dependencies will:
1. remove existing security warnings, e.g.:  [Dependabot alert #27](https://github.com/apache/incubator-teaclave-trustzone-sdk/security/dependabot/27) , and other warnings of `rustls`, `webpki`, and `ring`.  
2. demonstrate fully updated TLS examples

## Key Improvements in Rustls 0.23+  
Rustls now supports:  
- **[CryptoProvider](https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html)**  
- **[TimeProvider](https://docs.rs/rustls/latest/rustls/time_provider/trait.TimeProvider.html)**  

This allows us to integrate directly with `optee-utee` APIs without manually porting entire crates.  
After customizing the rng and time providers, Rustls can run in OP-TEE `std` TAs.  (`no-std` support is in future plan but not included here.)  

- **TimeProvider**: Straightforward to implement, please see code in this PR.  
- **CryptoProvider**: More involved, detailed below.  

## Crypto Provider  

Available Rustls crypto providers:  <https://github.com/rustls/rustls?tab=readme-ov-file#cryptography-providers>  
Of course we can implement and maintain our own crypto provider for OP-TEE, but if there are existing providers available, it’s better to reuse them. We selected `ring` and `rustls-rustcrypto` as our candidates.

### 1. `ring` provider  
- Built into Rustls (via the `ring` feature).  (https://github.com/rustls/rustls/tree/main/rustls/src/crypto/ring)
- Dependency chain: `ring-provider-in-rustls` → `ring` → `getrandom`.  
- Problem: `ring`’s `SystemRandom` is not OP-TEE–compatible. (https://github.com/briansmith/ring/blob/main/src/rand.rs#L165)  
  Would require changes to both `getrandom` **and** `ring` (or `ring-provider-in-rustls`).  

### 2. `rustls-rustcrypto` provider  
- Third-party provider.  (https://github.com/RustCrypto/rustls-rustcrypto)
- Dependency chain: `rustls-rustcrypto` → `rand_core` → `getrandom`.  
- Only requires `getrandom` modifications. No changes to the provider code itself.  

Our decision is, port `getrandom` to OP-TEE, then use `rustls-rustcrypto` as the provider.  

## Getrandom Porting  

- `getrandom` is a hardware RNG crate with multiple backends.  
- We have added an OP-TEE target and backend: 
    - Since `rustls-rustcrypto` depends on `getrandom 0.2` (which is in maintenance mode), we maintain an OP-TEE-enabled `getrandom` in **`incubator-teaclave-crates`** (currently in my fork: `git = "https://github.com/DemesneGH/incubator-teaclave-crates.git"`, I will open a PR and modify this url after it merged).  
    - We plan to upstream OP-TEE support for `getrandom 0.3`.  
- Because the ported `getrandom` depends on `optee-utee` **0.4.0**, all other TLS TA modules also pin `optee-*` crates to 0.4.0 to avoid build inconsistencies.  

## Code Changes  

- New `tls_client` and `tls_server` examples using Rustls `0.23.12`. We chose this version because it is the one currently used by `rustls-rustcrypto`.
- New `rustls-provider` crate under `crates/`, used by TLS examples.  
- Updated outdated TLS server test certs and improved test script error checks.  
- Fixed 32-bit TLS TA build errors by disabling `CFG_TA_FLOAT_SUPPORT` (see commit message for details).  

## `no-std` TLS examples

They need the `no-std` Rustls and `no-std` crypto backend. AFAIK:
- Rustls now offers initial support for `no-std` ([release notes](https://github.com/rustls/rustls/releases/tag/v%2F0.23.0)), which requires further research for our usecase.  
- Several `rustcrypto` crates already support `no-std`.  
It is possible to have `no-std` TLS TAs, but this will require additional effort and investigation, so it remains a goal for future work.


examples/tls_client-rs/ta/src/main.rs
```diff
@@ -64,34 +63,35 @@ fn invoke_command(cmd_id: u32, _params: &mut Parameters) -> Result<()> {
...
+    let mut config = rustls::ClientConfig::builder_with_details(crypto_provider, time_provider)
+        .with_safe_default_protocol_versions()
+        .expect("inconsistent cipher-suite/versions selected")
         .with_root_certificates(root_store)
         .with_no_client_auth();
-    trace_println!("[+] config created");
 
-    let server_name = "google.com".try_into().unwrap();
+    // Allow using SSLKEYLOGFILE.
+    config.key_log = Arc::new(rustls::KeyLogFile::new());
```
m4sterchain: Any reason to allow KeyLogFile in TA?

examples/tls_client-rs/ta/Cargo.toml
```diff
@@ -17,38 +17,30 @@
...
 lto = false
 opt-level = 3
 
 [patch.crates-io]
-ring = { git = "https://github.com/DemesneGH/ring-optee.git", branch = "0.16.20-optee" }
-
-# Patch optee-utee for rustls
-[patch."https://github.com/apache/incubator-teaclave-trustzone-sdk.git"]
-optee-utee = { path = "../../../optee-utee" }
+getrandom = { git = "https://github.com/DemesneGH/incubator-teaclave-crates.git" }
```
m4sterchain: Can we depend on unpatched getrandom and leveraging its `custom` feature?
https://docs.rs/getrandom/0.2.16/getrandom/index.html#macros
If it works, we don't need to maintain the ported crates.

build_optee_libraries.sh
```diff
@@ -41,7 +41,7 @@ export CROSS_COMPILE64="aarch64-linux-gnu-"
 
 # build optee_os and optee_client for qemu_v8
 git clone https://github.com/OP-TEE/optee_os.git -b $OPTEE_VERSION $OPTEE_DIR/optee_os
-(cd $OPTEE_DIR/optee_os && make PLATFORM=vexpress-qemu_armv8a)
+(cd $OPTEE_DIR/optee_os && make PLATFORM=vexpress-qemu_armv8a CFG_TA_FLOAT_SUPPORT=n)
```
m4sterchain, line 47: I am not sure the reason for this change. 
But we have another `build_optee_libraries.sh` in `scripts/setup` for update, if this change is necessary.

examples/tls_server-rs/ta/src/main.rs
```diff
@@ -92,12 +94,23 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
 }
 
 pub fn new_tls_session(session_id: u32) {
```
m4sterchain: 1. This function should return a Result, as there are multiple failures inside its impl.
2. The unwrap should always be eliminated, in our codebase.
I suggest to run `cargo clippy` to help check and follow best Rust coding practices.

examples/tls_server-rs/ta/src/main.rs
```diff
@@ -131,36 +144,44 @@ pub fn do_tls_write(session_id: u32, buf: &mut [u8]) -> usize {
...
+    let crypto_provider = Arc::new(rustls_provider::optee_crypto_provider());
+
+    trace_println!("[+] Creating time provider");
+    let time_provider = Arc::new(rustls_provider::optee_time_provider());
+
+    let certs = load_certs().context("Failed to load certificates")?;
+    trace_println!("[+] Loaded {} certificates", certs.len());
+
+    let private_key = load_private_key().context("Failed to load private key")?;
+    trace_println!(
```
m4sterchain: Private data should never be printed inside Trusted Applications (TAs), especially in demonstration examples.

examples/tls_server-rs/ta/test-ca/ecdsa/ca.cert
```diff
@@ -1,12 +1,12 @@
...
-obja/vQ6UBixlRB5vDSG0UuukL4kzlyUKpHkwUcj
+MIIBvDCCAUKgAwIBAgIUcfobdnUAGg7+ZKWcoJDhubf76iEwCgYIKoZIzj0EAwIw
+FTETMBEGA1UEAwwKdGVzdHNlcnZlcjAeFw0yNTA4MTMwMzQ5MTFaFw0zNTA4MTEw
+MzQ5MTFaMBUxEzARBgNVBAMMCnRlc3RzZXJ2ZXIwdjAQBgcqhkjOPQIBBgUrgQQA
+IgNiAARuhizLUdjzt2IufRL1ELnTris+0kHEkc1WYS/vSaj+8dMRZ9/MW8ScsMjn
+zs0K6XMme4o3TuPgFm3MdiWlIwHBN4CjkPN3u0U3N3Lw1DHvl/oUMfQ/mgBt9/Cz
+osOnSx2jUzBRMB0GA1UdDgQWBBRIkFT2dS2wya07AoKzpoL0E6SO/zAfBgNVHSME
+GDAWgBRIkFT2dS2wya07AoKzpoL0E6SO/zAPBgNVHRMBAf8EBTADAQH/MAoGCCqG
+SM49BAMCA2gAMGUCMQCD4bitceb4NJ9Snf3X3Watg3YlWv5mmeT3K6khxEiypQOd
+/XJC+UvUisbD9srbWLUCMBydSCIzPsS4CGIWQ+lZY1+tqWQ6FVW9iTQ7pYgZSibD
```
m4sterchain, line 10: Is there any specific reason for changing the existing test certs/keys?
It’s good to add a script that explains how to generate these test-CA artifacts. However, for the existing artifacts, we should keep them unchanged and reuse them as a demonstration that the updated TLS example remains compatible.

examples/tls_client-rs/ta/src/main.rs
DemesneGH: Just keep sync with https://github.com/rustls/rustls/blob/v/0.23.12/examples/src/bin/simpleclient.rs to demonstrate the same code can run in TA.

examples/tls_client-rs/ta/Cargo.toml
DemesneGH: The `custom` feature works but need to implement the `register_custom_getrandom` in the root binary (https://docs.rs/getrandom/0.2.16/getrandom/macro.register_custom_getrandom.html#registering-a-custom-getrandom-implementation). It's not convenient for developers because they need to add the registration lines of code in their TAs. For long-term maintainability, I prefer adding an `optee` target to `getrandom` instead.

build_optee_libraries.sh
DemesneGH: Please see the commit message:
```
When building tls_client and tls_server 32bit TAs, this error occurred:
  multiple definition of `__aeabi_fcmple' (`__aeabi_fcmpeq' and others)
This means the __aeabi functions are defined both in Rustc
compiler_builtins and optee libutils.

Disable the CFG_TA_FLOAT_SUPPORT when building libutils is a easy
approach without porting the compiler_builtins.
```

Since disable the float support in compiler_builtins requires manual porting, I think the current code change is the easiest way to fix the 32bit TA building error. Seems our current TAs don't need the float support, it works for now. If there're new TAs need the float support, we can handle it then.

If it's acceptable I will update the one in `scripts/setup`.

examples/tls_server-rs/ta/test-ca/ecdsa/ca.cert
DemesneGH: The original test certs were generated at 2019 and expired at 2024, "Cert Expired" error occurs when test tls_server example.

examples/tls_client-rs/ta/src/main.rs
m4sterchain: In the context of a TA, we should be careful with this. The `KeyLogFile implementation opens a file based on the SSLKEYLOGFILE environment variable and writes keys into it`. I don’t think either the environment variable or file-writing behavior makes sense or is applicable inside a TA, despite that `If such a file cannot be opened, or cannot be written then this does nothing but logs errors at warning-level`.

https://docs.rs/rustls/latest/rustls/struct.KeyLogFile.html

examples/tls_server-rs/ta/test-ca/ecdsa/ca.cert
m4sterchain: Thanks for this information. Mark as resolved.

build_optee_libraries.sh
m4sterchain: Thanks for outlining the underlying issue, and I agree with the proposed actions. 
I suggest inlining the reason for this change directly as a comment in the script for future maintenance, rather than keeping it only in the commit message. This will make it clearer for anyone reading the build scripts later.

examples/tls_client-rs/ta/Cargo.toml
m4sterchain: I agree, and I also believe that adding an OP-TEE target to the upstream `getrandom` crate is the right long-term approach.

However, my point here is whether we really need to depend on a modified getrandom. If possible, I’d like to demonstrate to developers that they can customize the upstream getrandom crate directly with SDK's randomness (with just a few lines of code), without reviewing any patched code. 

Once the OP-TEE target is upstreamed and adopted by the crypto crates, we can seamlessly remove the `custom` feature requirement.

examples/tls_client-rs/ta/src/main.rs
DemesneGH: removed

build_optee_libraries.sh
DemesneGH: updated code comment and same in `scripts/setup`

examples/tls_server-rs/ta/src/main.rs
DemesneGH: Since there're some clippy error in other modules which should not be resolved in this PR, the tls TAs passed with `xargo clippy --target aarch64-unknown-optee -- -D clippy::all -A clippy::missing-safety-doc -A clippy::needless-return`
For the full `clippy` for our codebase, I've opened a issue for tracking: https://github.com/apache/incubator-teaclave-trustzone-sdk/issues/222

examples/tls_server-rs/ta/src/main.rs
DemesneGH: removed

examples/tls_client-rs/ta/Cargo.toml
DemesneGH: I think both approaches make sense. I'd prefer the current "port getrandom" approach because:
The migration from the current approach to a future upstreamed `getrandom` would require no code changes—developers would simply remove the patch section from `Cargo.toml`.

> If possible, I’d like to demonstrate to developers that they can customize the upstream getrandom crate directly with SDK's randomness (with just a few lines of code)

It's good to have such demonstration.
However, in the OP-TEE context, developers often rely on the same RNG from OP-TEE APIs, I think the need to implement their own RNG is minimal. If there're indeed need for this, we can open a PR then.

m4sterchain: I don’t see the necessity of using patched crates as a dependency.
One of the key values of the current TLS examples is that they don’t depend on patched crates tailored for OP-TEE. You’ve almost succeeded in removing all patched crates—why not make the final push to remove the last one?

crates/rustls_provider/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+use std::time::Duration;
+
+/// CryptoProvider from rustls-rustcrypto, with the rng backend for OP-TEE in getrandom crate
+pub fn optee_crypto_provider() -> CryptoProvider {
+    rustls_rustcrypto::provider()
+}
+
+/// Custom TimeProvider implementation using OP-TEE UTEE API
+#[derive(Debug)]
+pub struct OpteeTimeProvider;
```
m4sterchain: `ReeTimeProvider` might be better than `OpteeTimeProvider`, as it explicitly shows the source of time info. 
The security properties for the ReeTimeProvider should also be hilighted.
`In normal operation, the value returned should correspond to the real time, but it should not be considered as trusted, as it may be tampered by the user or the REE software.` 
ref: GPD_TEE_Internal_API_Specification

crates/rustls_provider/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+name = "rustls_provider"
+version = "0.1.0"
+authors = ["Teaclave Contributors <dev@teaclave.apache.org>"]
+license = "Apache-2.0"
+repository = "https://github.com/apache/incubator-teaclave-trustzone-sdk.git"
+description = "Custom Rustls providers for OP-TEE TrustZone SDK."
+edition = "2018"
+
+[dependencies]
+optee-utee = "0.4.0"
```
m4sterchain: We should have "0.5.0" published on crates.io, as it's been a while for the v0.5.0 Apache release.

crates/rustls_provider/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+        time.ree_time();
+
+        // Convert OP-TEE time to Unix timestamp
+        // OP-TEE time seconds field represents seconds since some epoch
+        // We need to treat it as Unix timestamp (seconds since Jan 1, 1970)
+        let seconds = time.seconds as u64;
+        let millis = time.millis as u64;
+
+        // Create UnixTime from seconds and milliseconds
+        let total_millis = seconds * 1000 + millis;
```
m4sterchain: Because the Ree time can be manipulated, operations on it may overflow. Although the exact consequences are unclear, it would be safer to return None in such cases.

examples/tls_server-rs/ta/src/main.rs
```diff
@@ -63,104 +65,183 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
...
             trace_println!("[+] new_tls_session");
-            new_tls_session(session_id);
-            Ok(())
+            match new_tls_session(session_id) {
+                Ok(_) => Ok(()),
+                Err(e) => {
+                    trace_println!("[-] Failed to create TLS session: {:?}", e);
+                    Err(Error::new(ErrorKind::Generic))
+                }
+            }
```
ivila: I think we would better use `map_err` here?

examples/tls_server-rs/ta/src/main.rs
```diff
@@ -63,104 +65,183 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
...
             trace_println!("[+] do_tls_read");
-            do_tls_read(session_id, buffer);
-            Ok(())
+            match do_tls_read(session_id, buffer) {
+                Ok(_) => Ok(()),
+                Err(e) => {
+                    trace_println!("[-] Failed to read TLS data: {:?}", e);
+                    Err(Error::new(ErrorKind::Generic))
+                }
+            }
```
ivila: Suggest using `map_err` here too.

examples/tls_server-rs/ta/src/main.rs
```diff
@@ -63,104 +65,183 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
...
         Command::CloseTlsSession => {
             trace_println!("[+] close_tls_session");
-            close_tls_session(session_id);
-            Ok(())
+            match close_tls_session(session_id) {
+                Ok(_) => Ok(()),
+                Err(e) => {
+                    trace_println!("[-] Failed to close TLS session: {:?}", e);
+                    Err(Error::new(ErrorKind::Generic))
+                }
```
ivila: Suggest `map_err` here too.

crates/rustls_provider/Cargo.toml
DemesneGH: - This file is outdated and we use the relative path same as other examples
- BTW I don't know the current status of 0.5.0 release, have the all votes passed?

examples/tls_server-rs/ta/src/main.rs
```diff
@@ -63,104 +65,183 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
...
+            match new_tls_session(session_id) {
+                Ok(_) => Ok(()),
+                Err(e) => {
+                    trace_println!("[-] Failed to create TLS session: {:?}", e);
+                    Err(Error::new(ErrorKind::Generic))
+                }
+            }
         }
         Command::DoTlsRead => {
             let mut p1 = unsafe { params.1.as_memref().unwrap() };
```
ivila: Can we remove the `unwrap` here?

examples/tls_server-rs/ta/src/main.rs
```diff
@@ -63,104 +65,183 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
...
+                Err(e) => {
+                    trace_println!("[-] Failed to read TLS data: {:?}", e);
+                    Err(Error::new(ErrorKind::Generic))
+                }
+            }
         }
         Command::DoTlsWrite => {
             trace_println!("[+] do_tls_write");
             let mut p1 = unsafe { params.1.as_memref().unwrap() };
             let mut p2 = unsafe { params.2.as_value().unwrap() };
```
ivila: Can we remove the `unwrap` here?

crates/rustls_provider/Cargo.toml
m4sterchain: There are a few ways to check the release status:
	•	release page on GitHub: https://github.com/apache/incubator-teaclave-trustzone-sdk/releases/tag/v0.5.0
	•	Apache Teaclave website download page: https://teaclave.apache.org/download

For Apache vote tracking, you can subscribe or search the general@incubator.apache.org mailing list.

crates/rustls_provider/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+rustls-rustcrypto = "0.0.2-alpha"
+getrandom = "0.2"
+
+# Pin these crates for compatibility with our Rustc version nightly-2024-05-15
+base64ct = "=1.6.0"
+ed25519-dalek = "=2.1.0"
+
+[patch.crates-io]
+# For getrandom 0.2, we add the OP-TEE backend and maintain in teaclave crates
+getrandom = { git = "https://github.com/apache/incubator-teaclave-crates.git" }
```
m4sterchain: Can we also remove this patch? I suppose the patch (maybe even the dependency) is not required in this context when the customized feature is enabled.

examples/tls_client-rs/ta/src/main.rs
```diff
@@ -17,18 +17,28 @@
...
 
+// Register the custom getrandom implementation.
+//
+// In getrandom 0.2 there is no built-in OP-TEE target, so we rely on the
+// `custom` feature to provide an OP-TEE RNG.
+// Reference: https://docs.rs/getrandom/0.2.16/getrandom/macro.register_custom_getrandom.html
+//
+// For this example, the shared `optee_getrandom` function is defined in the
+// `rustls_provider` crate and registered here.
+getrandom::register_custom_getrandom!(rustls_provider::optee_getrandom);
```
m4sterchain, line 40: Can we simply use `getrandom::register_custom_getrandom!(optee_utee::Random::generate)` directly?  
If so, we wouldn’t need the extra wrapping logic in `rustls_provider::optee_getrandom`. 
The intent (using optee_utee's randomness) is already clear and self-explanatory.

We might also update the same pattern in tls_server-rs for consistency.

crates/rustls_provider/Cargo.toml
DemesneGH: v0.5.0 has been published: https://crates.io/search?q=optee

crates/rustls_provider/Cargo.toml
DemesneGH: Sorry I forgot to delete this, will update.

examples/tls_client-rs/ta/src/main.rs
DemesneGH: We can't, because the required custom fn has different signature with `optee_utee::Random::generate` (`pub fn generate(res_buffer: &mut [u8])`).
It requires: `pub fn custom_rng(buf: &mut [u8]) -> Result<(), getrandom::Error>`

m4sterchain: I just notice the difference about the return types. If they are not compatible, the wrapper makes sense to me now.

PR #226, opened 2025-08-27
ci: cleanup unused packages for OP-TEE workflows

DemesneGH:
- Remove unused packages from the host machine, freeing approximately 15 GB of disk space.

- Build directly on the host machine instead of using a container. Seems the container only installed OP-TEE dependencies https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/main/Dockerfile, which can now be handled directly on the host. Running everything on the host avoids unnecessary switching between host and container environments.

Since the CI is not triggered for PRs, you can see it passed on my fork: https://github.com/DemesneGH/rust-optee-trustzone-sdk


ivila (approved):
LGTM

PR #232, opened 2025-09-19
optee-utee: add std feature

DemesneGH:
This PR is the first step toward enabling Clippy checks on the `optee-utee` crates and unifying our `cfg` usage to better support both `std` and `no-std`.

#### Motivation

Previously, we relied on `target_os = "optee"` to distinguish between `std` and `no-std` builds:
- `aarch64-unknown-linux` → used for `no-std` TAs  
- `aarch64-unknown-optee` (local customized) → used for `std` TAs  

Because the `optee` target is not upstream target, clippy raises errors:
```
unexpected cfg condition value: optee 
--> optee-utee-sys/src/lib.rs:56:11 
| 56 | #[cfg(not(target_os = "optee"))]
```

Since the real distinction we care about is `std` vs. `no-std`, this PR replaces the `target_os` check with a `std` feature on `optee-utee`.

#### Advantages

- Defining a `std` feature in a Rust crate is a widely adopted approach (e.g., in crates `rand`, `chrono`). For our SDK, we stay `no-std` by default, with `std` as an opt-in feature.  
- Once the OP-TEE targets are upstreamed, both `std` and `no-std` will share the same `optee` target. Checking `target_os` will no longer be meaningful.  
- This aligns with the approach already used in [`tcp_client-rs` and `udp_client-rs`](https://github.com/apache/teaclave-trustzone-sdk/pull/224), giving us a consistent pattern across the SDK.  

#### Changes

- `optee-utee`, `optee-utee-sys`, `optee-utee-macros`: define the `std` feature.
- Common examples (support both `std` and `no-std`): updated `main.rs`, `Cargo.toml`, and `Makefile`.  
  - Note: I found the common examples define `#![no_std]` in their `main.rs` which doesn't enable `std` build. Changed to `#![cfg_attr(not(feature = "std"), no_std)]`
- `std`-only examples: updated `Cargo.toml`.  
- `no-std`-only examples: unchanged.  
- `examples/*/ta/Makefile`: unified `RUSTFLAGS := -C panic=abort` for both `std` and `no-std`.  

-------
#### Clippy Enforcement Plan

- [DONE] All examples: enable default Clippy + compiler checks ([https://github.com/apache/teaclave-trustzone-sdk/pull/224](https://github.com/apache/teaclave-trustzone-sdk/pull/224)).
- [DONE] All examples: enable advanced panic checks (https://github.com/apache/teaclave-trustzone-sdk/pull/229).
- [WIP] optee-* crates: enable default Clippy checks.
  - add `std` feature in optee-utee (this PR)
  - [TODO] fix other clippy errors in optee-*
- [TODO]optee-* crates: enable advanced Clippy checks.


examples/acipher-rs/ta/Makefile
```diff
@@ -24,23 +24,24 @@ OBJCOPY := $(CROSS_COMPILE)objcopy
...
-RUSTFLAGS := $(if $(STD),,-C panic=abort)
+# Set panic=abort for std and no-std
+RUSTFLAGS := -C panic=abort
 
 TA_SIGN_KEY ?= $(TA_DEV_KIT_DIR)/keys/default_ta.pem
 SIGN := $(TA_DEV_KIT_DIR)/scripts/sign_encrypt.py
 OUT_DIR := $(CURDIR)/target/$(TARGET)/release
 
 BUILDER ?= $(if $(STD),xargo,cargo)
+FEATURES := $(if $(STD),--features std,)
```
m4sterchain: We should avoid using STD env in the TA Makefile
- in legacy building system, the STD is a env variable used for setup the target envs.
- in docker building system, there is no STD env variable exist, as we use env files to setup the target envs.  
The design principle for the makefile is using the caller provided variable, e.g. passing caller provided FEATURES to cargo, instead of defining FEATURES conditionally; similar as TARGET, CROSS_COMPILE.

m4sterchain (commented):
I don't see changes to the top level makefiles.
- e.g. main/examples/Makefile, etc.
I suggest leverage them for the feature setup.

DemesneGH:
As discussed, I will test the new change on the dev docker, this PR will merged after the tests passed.

BTW I propose to add the CI jobs for testing on dev docker after this PR. Makefiles reorganization is planned after the CI enabled.

DemesneGH:
Since no followup comments, merging now.

PR #235, opened 2025-09-26
optee-*: clippy default enforcement in CI

DemesneGH:
This PR enables default Clippy and Rust compiler checks `cargo clippy -- -D warnings` for `optee-utee` and `optee-teec` in our CI.

It also fixes Clippy-reported issues, including `match-like-matches-macro`, `needless-borrowed-reference`, `not-unsafe-ptr-arg-deref`, `needless_return`, `unnecessary_cast`, and so on.

Some `#[allow(...)]` attributes are retained:
- `#[allow(non_camel_case_types)]` in `optee-*-sys`, which defines raw C structures.
- `#[allow(non_snake_case)]` in `optee-utee-mock`.

-------
#### Clippy Enforcement Plan

- [DONE] All examples: enable default Clippy + compiler checks ([https://github.com/apache/teaclave-trustzone-sdk/pull/224](https://github.com/apache/teaclave-trustzone-sdk/pull/224)).
- [DONE] All examples: enable advanced panic checks (https://github.com/apache/teaclave-trustzone-sdk/pull/229).
- [WIP] optee-* crates: enable default Clippy checks.
  - [DONE] add `std` feature in optee-utee (https://github.com/apache/teaclave-trustzone-sdk/pull/232)
  - fix other clippy errors in optee-* (this PR)
- [ADVANCED] optee-* crates: enable advanced Clippy checks.

### DIFF
--- a/optee-teec/macros/src/lib.rs
+++ b/optee-teec/macros/src/lib.rs
@@ -125,7 +125,7 @@ pub fn plugin_invoke(_args: TokenStream, input: TokenStream) -> TokenStream {
     quote!(
         /// # Safety
         ///
-        /// The `_plugin_invoke` function is the `extern "C"` entrypoint called by OP-TEE OS.  
+        /// The `_plugin_invoke` function is the `extern "C"` entrypoint called by OP-TEE OS.
         /// This SDK allows developers to implement the inner logic for a Normal World plugin in Rust.
         /// More about plugins:
         /// https://optee.readthedocs.io/en/latest/architecture/globalplatform_api.html#loadable-plugins-framework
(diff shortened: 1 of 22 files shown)


optee-teec/macros/src/lib.rs
```diff
@@ -129,11 +123,18 @@ pub fn plugin_invoke(_args: TokenStream, input: TokenStream) -> TokenStream {
...
+        /// 
+        /// This function is unsafe because it:
+        /// - Dereferences raw pointers `data` and `out_len`
+        /// - Creates a mutable slice from raw parts using `data` and `in_len`
+        /// - Copies data to the raw pointer `data`
+        /// 
+        /// The caller must ensure that:
+        /// - `data` points to valid memory for at least `in_len` bytes
+        /// - `out_len` points to valid memory for a `u32`
+        /// - `data` has enough capacity for the output data
```
m4sterchain: Security assumptions need to be discussed and highlighted at least in the follow scenarios
- Whether `data` can be null, or `in_len` can be 0
- Whether `outslice` and data can have overlap
- Whether `outslice.len()` can be greater than `in_len`

optee-teec/macros/src/lib.rs
```diff
@@ -143,16 +165,14 @@ pub fn plugin_invoke(_args: TokenStream, input: TokenStream) -> TokenStream {
             fn inner(#params) -> optee_teec::Result<()> {
                 #f_block
             }
-            let mut inbuf = unsafe { std::slice::from_raw_parts_mut(data, in_len as usize) };
+            let mut inbuf = std::slice::from_raw_parts_mut(data, in_len as usize);
```
m4sterchain: 1. **Null Check for `data`**  
   Add an explicit null check for the `data` pointer before calling [`from_raw_parts_mut`](https://doc.rust-lang.org/std/slice/fn.from_raw_parts_mut.html#safety).  
   The current implementation directly calls `from_raw_parts_mut(data, in_len as usize)`, which is undefined behavior if `data` is null—even when `in_len == 0`.

2. **Clarify Contradictory Documentation**  
   The function-level comments contain a small inconsistency:  
   - It states that a **valid empty call** allows `data = null` when `in_len = 0`.  
   - It also states that **invalid pointers** include `data = null`.  
   Please clarify this to avoid confusion, I suggest our ffi logic should handle such case as described in 1.

3. **Purpose of `get_out_slice` Copy**  
   The FFI layer wraps the `data` buffer with `params` and passes it to `inner(&mut params)`, which already allows reading/writing in place.  
   After `inner` returns, the FFI layer calls `params.get_out_slice()` and copies the output back to `data`.  
   - If `params` operates directly on the same buffer, what is the purpose of this additional copy?  
   - Could you please confirm if this is redundant writes or intentionally for some corner case?

4. **Handling of `out_len` on Insufficient Buffer**  
   In some C APIs that use an in/out buffer, `out_len` is used to return the *required* size when the provided buffer is too small.  
   It’s not clear whether this convention applies here.  
   - If so, the current design seems to ignore this scenario.  
   - If not, consider clarifying this in the documentation to avoid ambiguity.

optee-teec/src/extension.rs
```diff
@@ -57,7 +57,8 @@ impl<'a> PluginParameters<'a> {
     pub fn set_buf_from_slice(&mut self, sendslice: &[u8]) -> Result<()> {
         if self.inout.len() < sendslice.len() {
             println!("Overflow: Input length is less than output length");
-            return Err(Error::new(ErrorKind::Security));
+            self.outlen = sendslice.len();
```
m4sterchain: The latest API change has side affect that causes panic in certain calling sequence, which could be avoided with a more elegant impl. 
```
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_set_buf_from_slice_overflow() {
        let mut buffer = [0u8; 8];
        let mut params = PluginParameters::new(1, 2, &mut buffer);

        let data = [1u8; 16];
        let result = params.set_buf_from_slice(&data);
        let s = params.get_out_slice();
        assert!(result.is_err());
    }
}

thread 'tests::test_set_buf_from_slice_overflow' panicked at ...:
range end index 16 out of range for slice of length 8
```

PR #236, opened 2025-09-30
ci: test build on dev docker

DemesneGH:
Add CI jobs:
- `test-nostd-build-on-dev-docker`
- `test-std-build-on-dev-docker`

to ensure our changes don't break the building process on dev docker.

Note: the `ENTRYPOINT` defined in our Dockerfile is not applied when GitHub Actions runs jobs with container:, so the environment setup logic is included explicitly in the workflow steps.


DemesneGH:
> Have you tried to use 'entrypoint.sh $cmd' before explicitly setup the envs? The entrypoint.sh can act as a wrapper to run commands, I am curious whether that helps for simplicity.

I've tried but didn't work:
```
      - name: Build no-std and common examples
        run: |
          /entrypoint.sh make
```
```
[CI/test-nostd-build-on-dev-docker]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /var/run/act/workflow/1.sh] user= workdir=
| Setting up bash profile environment...
| TA_DEV_KIT_DIR is not correctly defined
| make: *** [Makefile:56: all] Error 1
[CI/test-nostd-build-on-dev-docker]   ❌  Failure - Main Build no-std and common examples [137.391648ms]
```

The error can be reproduced by:
```
$ docker run -it \
  --name teaclave_dev_test_ci \
  -v $(pwd):/root/teaclave_sdk_src \
  -w /root/teaclave_sdk_src \
  teaclave/teaclave-trustzone-emulator-nostd-optee-4.7.0-expand-memory:latest

# ctrl + p + q to exit and keep the container running

$ docker exec -it teaclave_dev_test_ci bash -c "/entrypoint.sh make"
Setting up bash profile environment...
TA_DEV_KIT_DIR is not correctly defined
make: *** [Makefile:56: all] Error 1
```

Seems the `entrypoint.sh` doesn't setup the env in current shell, only writes to the bash profile.

m4sterchain:
The barrier might be GitHub Actions’ default use of `bash --noprofile --norc`. We can override this by specifying `shell: bash -l {0}` in the workflow step. 
I don’t have the test environment on hand, so please help verify whether this works as expected.

PR #237, opened 2025-10-13
dev-docker: use customized rust toolchain path

m4sterchain:
## Core Problem Solved
Made Docker build dependencies independent of $HOME directory to support CI/CD environments (e.g., GitHub Actions) where runtime home directory may differ from build-time home.

## Key Changes by Category
1. Centralized Rust Environment Configuration (scripts/setup/bootstrap_env)
Added: RUSTUP_HOME=${TEACLAVE_TOOLCHAIN_BASE}/rustup
Added: CARGO_HOME=${TEACLAVE_TOOLCHAIN_BASE}/cargo
Removed: Default OPTEE_VERSION value (moved to Dockerfile for explicit control)
Impact: Single source of truth for Rust toolchain paths, used by both setup and runtime
2. Fixed Dockerfile Build Process (Dockerfile.dev)
Added: ENV OPTEE_VERSION=4.7.0 for explicit version control
Changed: RUN commands use bootstrap_env and ${CARGO_HOME}/env instead of $HOME/.cargo/env
Changed: Export RUST_STD_DIR to environment file instead of .profile
Impact: Build process independent of $HOME, version explicitly controlled
3. Streamlined Container Entrypoint (scripts/entrypoint.sh)
Removed: Duplicate environment loading (was loaded in entrypoint AND in .profile)
Removed: Direct sourcing of $HOME/.cargo/env
Changed: Command execution uses /bin/bash -lc "$*" for proper argument handling
Impact: Environment loads once, cleaner command execution, no duplication

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -19,7 +19,6 @@ FROM ubuntu:24.04 AS no-std-build-env
 
 # Base environment
 ENV TEACLAVE_TOOLCHAIN_BASE=/opt/teaclave
-ENV OPTEE_VERSION=4.7.0
 ENV DEBIAN_FRONTEND=noninteractive
 
 # Copy scripts into image


Dockerfile.dev
```diff
@@ -19,6 +19,7 @@ FROM ubuntu:24.04 AS no-std-build-env
 
 # Base environment
 ENV TEACLAVE_TOOLCHAIN_BASE=/opt/teaclave
+ENV OPTEE_VERSION=4.7.0
```
DemesneGH: The ENV is already set in line 31-32, used by https://github.com/apache/teaclave-trustzone-sdk/blob/main/scripts/release/build_dev_docker.sh which reads the optee version from `optee-version.txt` and build the image. It keeps the consistency (the only source of optee version is the txt) and avoids manually edit the docker file for each optee release.

FYI the other code which reads the optee version is: https://github.com/apache/teaclave-trustzone-sdk/blob/main/build_optee_libraries.sh#L22

Please consider the usecase and decide if this line should be removed.

m4sterchain: > The ENV is already set in line 31-32, used by https://github.com/apache/teaclave-trustzone-sdk/blob/main/scripts/release/build_dev_docker.sh which reads the optee version from `optee-version.txt` and build the image. It keeps the consistency (the only source of optee version is the txt) and avoids manually edit the docker file for each optee release.
> 
> FYI the other code which reads the optee version is: https://github.com/apache/teaclave-trustzone-sdk/blob/main/build_optee_libraries.sh#L22
> 
> Please consider the usecase and decide if this line should be removed.

Nice catch. The original purpose of this line add is as a replacement of `export OPTEE_VERSION="${OPTEE_VERSION:-4.5.0}"` in [scripts/setup/bootstrap_env](https://github.com/apache/teaclave-trustzone-sdk/pull/237/files#diff-7b086229db005c0b13168adb3f614c922ca3f0ce4e2c4aa740d39e4ebfc5840c).

I just noticed that OPTEE_VERSION has already been enforced as a required build ARG (LINE31 in this file). 
So the replacement is unnecessary, we only need to remove that line, to avoid fallback to 4.5.0.

PR #243, opened 2025-10-21
ci: run test-in-optee-repo every day to detect issues earlier

ivila:


DemesneGH:
If we have the daily building check do we need the task for each merging into main?

PR #245, opened 2025-10-23
experimental: add cargo-optee

DemesneGH:
As discussed in https://github.com/apache/teaclave-trustzone-sdk/pull/244, I'm trying to remove Makefiles and provide a Cargo-based tool for building OP-TEE apps.

This is an initial version of the `cargo-optee` tool which enables building TAs in `no-std` and `std`:
```
$ cargo-optee build -h
Build a Trusted Application (TA)

Usage: cargo-optee build [OPTIONS] --ta_dev_kit_dir <TA_DEV_KIT_DIR> <TYPE>

Arguments:
  <TYPE>  Type of build target (currently only 'ta' is supported)

Options:
      --path <PATH>
          Path to the TA directory (default: current directory) [default: .]
      --arch <ARCH>
          Target architecture: aarch64 or arm (default: aarch64) [default: aarch64]
      --std
          Enable std feature for the TA
      --ta_dev_kit_dir <TA_DEV_KIT_DIR>
          Path to the TA dev kit directory (mandatory)
      --signing_key <SIGNING_KEY>
          Path to the TA signing key (default: $(TA_DEV_KIT_DIR)/keys/default_ta.pem)
      --uuid_path <UUID_PATH>
          Path to the UUID file (default: ../uuid.txt) [default: ../uuid.txt]
      --debug
          Build in debug mode (default is release)
  -h, --help
          Print help
```
build the cargo-optee tool:
```
cd cargo-optee && cargo build --release
```
building no-std TA:
```
cargo-optee build ta --path /teaclave/examples/hello_world-rs/ta --ta_dev_kit_dir /teaclave/optee/optee_os/out/arm-plat-vexpress/export-ta_arm64
```
building std TA:
```
cargo-optee build ta --path /teaclave/examples/hello_world-rs/ta --ta_dev_kit_dir /teaclave/optee/optee_os/out/arm-plat-vexpress/export-ta_arm64 --std
```

The goal is to make TA building simple: once the environment has the Rust and GCC toolchains installed, developers can build a TA directly using this tool.
It does not rely on global environment variables, only reads parameters explicitly passed. This design makes it more flexible across different development environments and provides clearer control over variable usage. The only mandatory parameter is `TA_DEV_KIT_DIR`; all other parameters have default values. 

However, `TA_DEV_KIT_DIR` is a low-level configuration detail. For our dev/emulation environments, we can provide a helper script in the future, to set up all required OP-TEE libraries and pass `TA_DEV_KIT_DIR` to the tool.

There are still several areas that can be further refined:

* Remove all `Xargo.toml` files from the TA source code.
* Automatically download std dependencies when the `std` feature is enabled.
* Enable building of CAs.
* Improve **UUID** location. It’s shared between the host and TA, so it may make sense to define it under `proto/`. The host can read it from the source code, and `cargo-optee` can use it when building the TA. For now, it remains as the original `uuid.txt`.
* Consider adding configurable metadata in the TA’s `Cargo.toml` (similar to what [[cargo-sgx](https://github.com/automata-network/automata-sgx-sdk)](https://github.com/automata-network/automata-sgx-sdk) does with `[package.metadata.sgx]`).
* Provide a new setup script for the dev/emulation environments.

I’ll move this forward. Please feel free to try out the tool and share any feedback about this, thanks!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,7 +95,7 @@ jobs:
           switch_config --ta no-std/arm32 && \
           switch_config --host arm32 && \
           source ${TEACLAVE_TOOLCHAIN_BASE}/environment && \
-          ./ci/build.sh --ta arm --ca arm
+          ./ci/build.sh --ta arm --host arm
 
   # Test cargo-optee build std examples - aarch64
   test-cargo-optee-std-aarch64:
@@ -145,7 +145,7 @@ jobs:
           switch_config --ta std/arm32 && \
           switch_config --host arm32 && \
           source ${TEACLAVE_TOOLCHAIN_BASE}/environment && \
-          ./ci/build.sh --ta arm --ca arm --std
+          ./ci/build.sh --ta arm --host arm --std
 
   license:
     runs-on: ubuntu-latest
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -30,25 +30,31 @@ Required environment variables:
 
 Options:
   --ta <arch>             TA architecture: aarch64 or arm (default: aarch64)
-  --ca <arch>             CA architecture: aarch64 or arm (default: aarch64)
+  --host <arch>           Host architecture for CA and plugins: aarch64 or arm (default: aarch64)
   --std                   Install with std support (default: no-std)
+  --ta-install-dir <path> TA installation directory (default: ./tests/shared)
+  --ca-install-dir <path> CA installation directory (default: ./tests/shared)
+  --plugin-install-dir <path> Plugin installation directory (default: ./tests/shared)
   --help                  Show this help message
 
 Examples:
   # Install for aarch64 in no-std mode
   TA_DEV_KIT_DIR=/path/to/export-ta_arm64 OPTEE_CLIENT_EXPORT=/path/to/export ./build.sh
 
-  # Install for ARM32 in std mode
-  TA_DEV_KIT_DIR=/path/to/export-ta_arm32 OPTEE_CLIENT_EXPORT=/path/to/export ./build.sh --ta arm --ca arm --std
+  # Install for ARM32 in std mode with custom directories
+  TA_DEV_KIT_DIR=/path/to/export-ta_arm32 OPTEE_CLIENT_EXPORT=/path/to/export ./build.sh --ta arm --host arm --std --ta-install-dir /target/lib/optee_armtz --ca-install-dir /target/usr/bin
 
-Note: Binaries are installed to './tests/shared' directory for use by test scripts.
+Note: Binaries are installed to './tests/shared' directory by default.
 EOF
 }
 
 # Parse command line arguments
 ARCH_TA="aarch64"  # Default: aarch64
-ARCH_CA="aarch64"  # Default: aarch64
+ARCH_HOST="aarch64"  # Default: aarch64
 STD=""             # Default: empty (no-std)
+TA_INSTALL_DIR=""  # Default: will be set to ./tests/shared if not specified
+CA_INSTALL_DIR=""  # Default: will be set to ./tests/shared if not specified
+PLUGIN_INSTALL_DIR=""  # Default: will be set to ./tests/shared if not specified
 
 # Parse arguments (support both positional and flag-style)
 while [[ $# -gt 0 ]]; do
@@ -61,22 +67,34 @@ while [[ $# -gt 0 ]]; do
             ARCH_TA="$2"
             shift 2
             ;;
-        --ca)
-            ARCH_CA="$2"
+        --host)
+            ARCH_HOST="$2"
             shift 2
             ;;
         --std)
             STD="std"
             shift
             ;;
+        --ta-install-dir)
+            TA_INSTALL_DIR="$2"
+            shift 2
+            ;;
+        --ca-install-dir)
+            CA_INSTALL_DIR="$2"
+            shift 2
+            ;;
+        --plugin-install-dir)
+            PLUGIN_INSTALL_DIR="$2"
+            shift 2
+            ;;
         *)
             # Positional arguments (backward compatibility)
             if [[ -z "${ARCH_TA_SET:-}" ]]; then
                 ARCH_TA="$1"
                 ARCH_TA_SET=1
-            elif [[ -z "${ARCH_CA_SET:-}" ]]; then
-                ARCH_CA="$1"
-                ARCH_CA_SET=1
+            elif [[ -z "${ARCH_HOST_SET:-}" ]]; then
+                ARCH_HOST="$1"
+                ARCH_HOST_SET=1
             elif [[ "$1" == "std" ]]; then
                 STD="std"
             fi
@@ -91,8 +109,8 @@ if [[ "$ARCH_TA" != "aarch64" && "$ARCH_TA" != "arm" ]]; then
     exit 1
 fi
 
-if [[ "$ARCH_CA" != "aarch64" && "$ARCH_CA" != "arm" ]]; then
-    echo "Error: ARCH_CA must be 'aarch64' or 'arm'"
+if [[ "$ARCH_HOST" != "aarch64" && "$ARCH_HOST" != "arm" ]]; then
+    echo "Error: ARCH_HOST must be 'aarch64' or 'arm'"
     exit 1
 fi
 
@@ -110,7 +128,7 @@ fi
 echo "==========================================="
 echo "Installing with configuration:"
 echo "  ARCH_TA: $ARCH_TA"
-echo "  ARCH_CA: $ARCH_CA"
+echo "  ARCH_HOST: $ARCH_HOST"
 echo "  STD: ${STD:-no-std}"
 echo "  TA_DEV_KIT_DIR: $TA_DEV_KIT_DIR"
 echo "  OPTEE_CLIENT_EXPORT: $OPTEE_CLIENT_EXPORT"
@@ -141,10 +159,29 @@ fi
 echo ""
 echo "Step 2: Installing all examples..."
 
-# Create shared directory for binaries (used by tests)
-SHARED_DIR="$(pwd)/tests/shared"
-mkdir -p "$SHARED_DIR"
-echo "Installing binaries to: $SHARED_DIR"
+# Set up installation directories
+# Each directory defaults to ./tests/shared if not specified
+if [ -z "$TA_INSTALL_DIR" ]; then
+    TA_INSTALL_DIR="$(pwd)/tests/shared"
+fi
+
+if [ -z "$CA_INSTALL_DIR" ]; then
+    CA_INSTALL_DIR="$(pwd)/tests/shared"
+fi
+
+if [ -z "$PLUGIN_INSTALL_DIR" ]; then
+    PLUGIN_INSTALL_DIR="$(pwd)/tests/shared"
+fi
+
+# Create all directories
+mkdir -p "$TA_INSTALL_DIR"
+mkdir -p "$CA_INSTALL_DIR"
+mkdir -p "$PLUGIN_INSTALL_DIR"
+
+echo "Installing binaries to:"
+echo "  TAs: $TA_INSTALL_DIR"
+echo "  CAs: $CA_INSTALL_DIR"
+echo "  Plugins: $PLUGIN_INSTALL_DIR"
 
 EXAMPLES_DIR="$(pwd)/examples"
 METADATA_JSON="$EXAMPLES_DIR/metadata.json"
@@ -255,7 +292,7 @@ for EXAMPLE_NAME in "${ALL_EXAMPLES[@]}"; do
             
             # Run cargo-optee install and capture both stdout and stderr
             if $CARGO_OPTEE install ta \
-                --target-dir "$SHARED_DIR" \
+                --target-dir "$TA_INSTALL_DIR" \
                 --ta-dev-kit-dir "$TA_DEV_KIT_DIR" \
                 --arch "$ARCH_TA" \
                 $TA_STD_FLAG; then
@@ -303,9 +340,9 @@ for EXAMPLE_NAME in "${ALL_EXAMPLES[@]}"; do
         cd "$CA_DIR_FULL_PATH"
         
         if $CARGO_OPTEE install ca \
-            --target-dir "$SHARED_DIR" \
+            --target-dir "$CA_INSTALL_DIR" \
             --optee-client-export "$OPTEE_CLIENT_EXPORT" \
-            --arch "$ARCH_CA"; then
+            --arch "$ARCH_HOST"; then
             echo "  ✓ CA installed successfully"
             # Clean up build artifacts
             $CARGO_OPTEE clean
@@ -349,9 +386,9 @@ for EXAMPLE_NAME in "${ALL_EXAMPLES[@]}"; do
         cd "$PLUGIN_DIR_FULL_PATH"
         
         if $CARGO_OPTEE install plugin \
-            --target-dir "$SHARED_DIR" \
+            --target-dir "$PLUGIN_INSTALL_DIR" \
             --optee-client-export "$OPTEE_CLIENT_EXPORT" \
-            --arch "$ARCH_CA"; then
+            --arch "$ARCH_HOST"; then
             echo "  ✓ Plugin installed successfully"
             # Clean up build artifacts
             $CARGO_OPTEE clean


cargo-optee/src/main.rs
```diff
@@ -0,0 +1,138 @@
...
+pub(crate) struct Cli {
+    #[clap(subcommand)]
+    cmd: Command,
+}
+
+#[derive(Debug, Subcommand)]
+enum Command {
+    /// Build a Trusted Application (TA)
+    #[clap(name = "build")]
+    Build {
```
ivila: Each build type uses its own set of options, with some overlap.
Making build_type a subcommand would help developers avoid irrelevant options.

For demo:
```rust
use clap::{Parser, Subcommand};
use std::path::PathBuf;

#[derive(Debug, Parser)]
struct BuildTypeCommonOptions {
    /// Path to the TA directory (default: current directory)
    #[arg(long = "path", default_value = ".")]
    path: PathBuf,

    /// Target architecture: aarch64 or arm (default: aarch64)
    #[arg(long = "arch", default_value = "aarch64")]
    arch: String,

    /// Path to the UUID file (default: ../uuid.txt)
    #[arg(long = "uuid_path", default_value = "../uuid.txt")]
    uuid_path: PathBuf,

    /// Build in debug mode (default is release)
    #[arg(long = "debug")]
    debug: bool,
}

#[derive(Debug, Subcommand)]
enum BuildCommand {
    TA {
        /// Enable std feature for the TA
        #[arg(long = "std")]
        std: bool,

        /// Path to the TA dev kit directory (mandatory)
        #[arg(long = "ta_dev_kit_dir", required = true)]
        ta_dev_kit_dir: PathBuf,

        /// Path to the TA signing key (default: $(TA_DEV_KIT_DIR)/keys/default_ta.pem)
        #[arg(long = "signing_key")]
        signing_key: Option<PathBuf>,

        #[command(flatten)]
        common: BuildTypeCommonOptions,
    },
}

#[derive(Debug, Subcommand)]
enum Command {
    /// Build a Trusted Application (TA)
    #[clap(name = "build")]
    #[command(subcommand)]
    Build(BuildCommand),
}

#[derive(Debug, Parser)]
#[clap(version = env!("CARGO_PKG_VERSION"))]
#[clap(about = "Build tool for OP-TEE Rust projects")]
pub(crate) struct Cli {
    #[clap(subcommand)]
    cmd: Command,
}


fn main() {
    Cli::parse();
}
```

Developers will run `${cmd} build ta --${options}` instead of `${cmd} build --build_ta=ta --${options}`, and they will get clearer help messages(only the relevant options involve)

image

cargo-optee/src/main.rs
```diff
@@ -0,0 +1,138 @@
...
+            path,
+            arch,
+            std,
+            ta_dev_kit_dir,
+            signing_key,
+            uuid_path,
+            debug,
+        } => {
+            // Validate build type
+            if build_type != "ta" {
```
ivila: Turning build_type into a subcommand helps eliminate the need for such comparisons.

cargo-optee/src/ta_builder.rs
```diff
@@ -0,0 +1,329 @@
...
+    );
+    bail!(
+        "{} failed with exit code: {:?}",
+        cmd_name,
+        output.status.code()
+    )
+}
+
+pub struct TaBuildConfig {
+    pub arch: String,            // "aarch64" or "arm"
```
ivila: Suggest using enum instead of String for `arch`, and derive `ValueEnum` for usage in clap.

cargo-optee/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+        #[arg(long = "optee_client_export", required = true)]
+        optee_client_export: PathBuf,
+
+        #[command(flatten)]
+        common: BuildTypeCommonOptions,
+    },
+}
+
+#[derive(Debug, Subcommand)]
+enum Command {
```
m4sterchain, line 275: Do we have a high level design about the cargo-optee?

What is the convention | requirement  between different roles?
- cargo-optee and underlining build system 
- cargo-optee and TA developer

Please refer to the following link as a reference. 
`https://github.com/automata-network/automata-sgx-sdk?tab=readme-ov-file#getting-started`

optee-utee-build/src/linker.rs
```diff
@@ -126,10 +126,11 @@ impl Linker {
...
     ) -> Result<(), Error> {
-        const ENV_TARGET_TA: &str = "TARGET_TA";
-        println!("cargo:rerun-if-env-changed={}", ENV_TARGET_TA);
+        // cargo passes TARGET as env to the build scripts
+        const ENV_TARGET: &str = "TARGET";
+        println!("cargo:rerun-if-env-changed={}", ENV_TARGET);
         let mut aarch64_flag = true;
-        match env::var(ENV_TARGET_TA) {
+        match env::var(ENV_TARGET) {
             Ok(ref v) if v == "arm-unknown-linux-gnueabihf" || v == "arm-unknown-optee" => {
```
m4sterchain, line 134: This function & inner logic is not straight forward and difficult to maintain.

Please leverage the Rust type system an use linear logic, instead of the boolean flag and nested if/else.
The convention here between src/ta.ld.S and created ta.lds is not straight forward.

It seems we can make it better by clearly handling Aarch64 and Arm32 separately and better leveraging the Type system to check the supported configurations.

cargo-optee/src/main.rs
DemesneGH: Updated in the new commit and `cargo-optee/README.md`

optee-utee-build/src/linker.rs
DemesneGH: I’ve checked whether the interfaces of `optee-utee-build` crate need any adjustments for the new building system.

After reviewing the interface layer (from cargo to the build script parameter passing), the build script mainly reads Cargo auto-set environment variables and the `TA_DEV_KIT` set by developer. I think this interface looks clear and does not require changes (see the appendix in `cargo-optee/README.md` for details).

Therefore, this is indeed an inner logic refinement issue. Let's create a separate issue to track this refactoring and resolve it in another PR.

cargo-optee/src/main.rs
m4sterchain: Thanks for the effort for showing the current design. 

```bash
cargo-optee build ta \
  --ta_dev_kit_dir <PATH> \
  [--path <PATH>] \
  [--arch aarch64|arm] \
  [--std] \
  [--signing_key <PATH>] \
  [--uuid_path <PATH>] \
  [--debug]
```

However, I don’t think this should be the ideal interface for TA developers.
I suggest we align and compare the design with existing cargo-based workflows such as cargo and cargo-sgx to ensure consistency and ease of use.

### Project-related configurations (for both TA and CA) should remain in Cargo.toml.
1. For example, `cargo-trustzone` should automatically infer whether to build with std based on the std feature declared in `Cargo.toml`, as you have proposed using the `std` feature for the std build in previous PR.
2. For other project-specific parameters, we can follow the convention used by `cargo-sgx`, which leverages `[package.metadata]` as the bridge between developers and the custom cargo toolchain.

### Toolchain configurations & Zero-argument usability
By default, running `cargo-trustzone build` should work without requiring additional parameters.
Since project-specific configurations are defined in `Cargo.toml`, the toolchain should have clear default discovery rules for common paths (e.g., how `ta_dev_kit_dir` is located, instead of a required parameter for each invocation of the build cli).
For example, the standard `cargo` toolchain automatically discovers the Rust sysroot and standard library paths by resolving from `rustc --print sysroot`, without needing users to specify them manually. Its behavior might also be influenced by environment variables such as RUSTUP_HOME and CARGO_HOME, which allow users to override toolchain locations in a standardized way.
Similarly, `cargo-trustzone` should support automatic discovery of the default `ta_dev_kit_dir` (e.g., from an environment variable, configuration file, or standard installation path) to provide a smooth and familiar developer experience consistent with the normal cargo workflow.

### Learn from Existing Design
Lastly, as I have suggested multiple times, please take a deeper look at the cargo-sgx tool and compare our current design against it. Many of the challenges we are encountering have already been addressed there — such as providing a developer-friendly interface, automatic build support for customized std environments, and seamless integration of loadable trusted components with untrusted applications. There may be additional insights and techniques we can learn from those design patterns. I strongly recommend gaining a solid understanding of what is already available before reinventing the wheel.

optee-utee-build/src/linker.rs
m4sterchain: Agree. We can open an issue to track this inner logic refinement, which seems like a good starting point for new committers to contribute.

cargo-optee/src/main.rs
DemesneGH: I've reviewed the `cargo-sgx` code snippet — although it reads metadata, it still accepts parameters via command line with default values. I assume you're suggesting we should move some params to `[package.metadata]` in `Cargo.toml` instead of passing them as command-line arguments?

I actually mentioned the idea in the initial message of this PR, but there are a few considerations. Let's go through each:

#### `arch` / `std`

We should allow building the same app for different arch and modes.
If these values are hardcoded in `Cargo.toml`, we lose the flexibility. For example, in CI:

> `cargo-trustzone` should automatically infer whether to build with `std` based on the feature declared in `Cargo.toml`.

But what should we write in the common examples’ `Cargo.toml`?
If we set `"no-std"`, how would we build it in `std` mode for CI? by modifying `Cargo.toml` each time? I don't have the answer.

Therefore, I believe **`arch`** and **`std`** should remain as **command-line parameters**, each with default values. 

#### `ta_dev_kit_dir`

Seems this might suitable for `Cargo.toml`, but it’s different for `arch` — for example:

```
optee_os/out/arm-plat-vexpress/export-ta_arm64
optee_os/out/arm-plat-vexpress/export-ta_arm32
```

If we put it in `Cargo.toml`, we’d have to edit that file each time we change the build architecture.
Maybe try this?
```
[metadata]
ta_dev_kit.64 = "xxxx_export-ta_arm64"
ta_dev_kit.32 = "xxxx_export-ta_arm32"
```
use the different entry decided by cmd param `arch`.

#### `uuid_path` (or `uuid`)

Instead of storing a `uuid_path`, it may make more sense to store a `uuid` string directly in the metadata. Since `uuid.txt` is both used by TA build tools and read by the proto (and imported by the CA when invoking the TA). We could try to write into the Cargo.toml, but to which Cargo.toml? Maybe we should have a workspace?
I’m still thinking about it.

#### signing_key

It makes sense to write into the Cargo.toml.

----

> and seamless integration of loadable trusted components with untrusted applications

Does it mean the plugin if in optee? I don't really get the point.


> Similarly, `cargo-trustzone` should support automatic discovery of the default `ta_dev_kit_dir` (e.g., from an environment variable, configuration file, or standard installation path) to provide a smooth and familiar developer experience consistent with the normal cargo workflow.

Consider the scenario:
- OP-TEE OS developers export the ta_dev_kit, and send it to the TA developer
- TA developer receives it and build the TA

For this tool, I want it less dependent with the environment, for production use cases or experienced developers, the TA developers typically only need a toolchain setup and already know the location of their ta_dev_kit; with those information they can build the TA directly. 

Additionally, I plan to provide a helper script for the dev/emulation env for easy use on qemuv8, together with preparing the OP-TEE libraries.

------
Please feel free to share your suggestions, thanks!

optee-utee-build/src/linker.rs
DemesneGH: tracked in https://github.com/apache/teaclave-trustzone-sdk/issues/250

cargo-optee/src/main.rs
m4sterchain: Before we move further, could you clarify what kind of user experience you envision for TA developers using the cargo-optee toolchain?

From my perspective, **the cargo-optee build command should provide a seamless interface aligned with the existing cargo build workflow**, either through transparent parameter passing or through complementary extensions. The key principle is to stay consistent with Cargo’s CLI conventions and user expectations.
It’s perfectly fine for the internal implementation to have a different interface, but here I’m mainly referring to the **user-facing design**.

Here I just expand one point for illustration.
Q: “But what should we write in the common examples’ Cargo.toml? If we set no-std, how would we build it in std mode for CI? By modifying Cargo.toml each time? I don't have the answer.”

My view: when designing the interface, we should ask — how do Rust developers normally address this problem (e.g., switching between no_std and std, or specifying architectures)? We should adopt the same mechanism.

For std / no_std switching, Rust developers use Cargo features declared in Cargo.toml, together with `--no-default-features` or `--features xxx` or nothing for `cargo build`. This convention can also be applied for our customized tooling.

The underlying build logic can follow Cargo’s native feature resolution model (via `cargo_metadata`), for example:
```
cargo (optee) build
cargo (optee) build --no-default-features
cargo (optee) build --no-default-features --features no_std // or std
```
The key concept here is transparent delegation to Cargo, cargo-optee does not need to reimplement or reinterpret feature logic, or use a reinvented `--std` parameter, but instead query and respect Cargo’s native feature resolution. This ensures consistent developer experience and predictable build behavior.

A similar comparison and concept mapping approach to existing Cargo mechanisms can be applied to other aspects as well, such as toolchain conventions and project conventions, to achieve a more consistent and intuitive CLI experience.

DemesneGH: I think `cargo-optee` provides the convenience for building TAs, not mandatory to keep consistency with the original cargo cmd parameters.

For example, the cargo-optee cmd:
```
cargo-optee build ta \
  --ta_dev_kit_dir /opt/optee/export-ta_arm64 \
  --path ./ta \
  --arch aarch64
```
will be converted to:
```
TA_DEV_KIT_DIR=/opt/optee/export-ta_arm64 \
RUSTFLAGS="-C panic=abort" \
cargo build --target aarch64-unknown-linux-gnu --release \
  --config target.aarch64-unknown-linux-gnu.linker="aarch64-linux-gnu-gcc"
```
for build.

`--arch` is not consistent with cargo's cmdline but it will configure the target for developer's convenience. Do you mean we should use the `--target` in `cargo-optee`?
If so, why not directly use cargo command, we don't need a wrapper.

`feature std` is similar. It's not only converted to the `cargo build --features std`, but also decide the `target` and whether we use `xargo`. I can follow the cargo's cmd format, like `cargo-optee build ta --features std`, but it doesn't mean the same as in cargo.

> From my perspective, **the cargo-optee build command should provide a seamless interface aligned with the existing cargo build workflow**,

I've checked other tools `cargo-sgx` `cargo-near` which also defines their cmd lines different with original cargo. Any reasons on why we should follow `seamless interface` principle?

m4sterchain: `cargo_metadata` retrieves meta info based on `cargo.toml` and command-line options in a programatic way, which does not mean convert to `cargo build --features std`. It is designated to used from within a cargo-* executable.  https://crates.io/crates/cargo_metadata

Here is a more concrete example to illustrate the difference.
- `cargo-optee build ta --std` means no matter what default feature developer defines in the Cargo.toml, the customized toolchain will build in using std library (new convention is introduced by cargo-optee). 
- `cargo optee build` (with other cargo compatible flags) means the the customized toolchain will respect what feature the developer enables in the Cargo.toml (std/no-std) and automatically handles internally (select xargo to build stdlib or not) based on the active feature. 

`Any reasons on why we should follow seamless interface principle?` 
There’s no strict requirement to maintain full consistency with Cargo’s original command-line parameters. However, I am just wondering if our goal should be to make OP-TEE development feel like a natural extension of the standard Cargo experience — a design I believe is achievable for our current use cases. To realize this, we can learn from existing customized toolchains, e.g. both tools you mentioned also leverage `cargo_metadata` to retrieve project setup information.

I will end this conversation here, as I believe I’ve explained my points clearly and don’t want to spend more time on further debate.

DemesneGH:
The new commit:

1. Read from Cargo.toml metadata: 
   - Enable reading build parameters from `[package.metadata.optee.*]` sections
   - Implement priority-based configuration: CLI arguments > metadata > default value
   - Update `supp_plugin-rs/*/Cargo.toml` files for demonstration
   - See updated `cargo-optee/README.md` for detailed usage instructions

2. Custom build env var and features:
   - Allow users to specify custom environment variables and features
   - These options are automatically appended to `cargo build` commands

The CI build script (`ci/build.sh`) passes successfully for the default aarch64 no-std configuration.

----
How to use this tool:
1. clone the code;
2. build the cargo-optee:
```
(cd cargo-optee && cargo build --release)
# Optional: add to PATH for easier usage
export PATH=/path/to/cargo-optee/target/release:$PATH
```
4. run `cargo-optee --help` to show the help messages
5. build the app: e.g. 
```
# with metadata the app can be built without params
cd examples/supp_plugin-rs/plugin && cargo-optee build plugin
# without metadata, (if you don't want to change the directory, you can specify `--manifest-path` like the original cargo)
cargo-optee build ta --manifest-path /teaclave/examples/hello_world-rs/ta/Cargo.toml  --ta-dev-kit-dir /path/to/optee_os/out/arm-plat-vexpress/export-ta_arm64
```

At this stage, we’ll focus on user-facing functionality and bug fixes. Please feel free to try out the tool and share your feedback, thanks!

DemesneGH:
@asir66 Could you help to try out the tool and share your feedback? Thanks!

cargo-optee/src/config.rs
```diff
@@ -0,0 +1,637 @@
...
+use std::path::{Path, PathBuf};
+
+use crate::common::Arch;
+
+/// Build configuration that can be discovered from proto metadata
+#[derive(Debug, Clone)]
+pub struct BuildConfig {
+    pub arch: Arch,
+    pub debug: bool,
+    pub std: bool,
```
asir66, line 30: There is an issue here: when --std is not provided, std defaults to false, so cmd_std becomes Some(false) instead of None.

asir66:
> @asir66 Thanks for your detailed and careful testing!
> 
> > I encountered an issue related to path resolution in `cargo-optee`. The behavior is inconsistent between `mnist-rs` and other examples such as `hello_world-rs` and `acipher-rs`. Since the root cause is not yet clear to me, I would like to ask for guidance or insights from the community.
> 
> The `mnist-rs` have the workspace in `mnist/ta/Cargo.toml`, could you help to check if the issue related with this?
> 
> > I would like to ask whether the project would benefit from having a Makefile under the examples/ directory to manage or build all examples together. Is this something we might need? Like [examples/Makefile](https://github.com/apache/teaclave-trustzone-sdk/blob/23a041fd658c99f542e3ba83331784fbe8453819/examples/Makefile)
> 
> The `ci/build.sh` performs as what original Makefile does, it reads the `examples/metadata.json` to get all app path of examples and build. Please also help to check if it enough for our use case?
> 
> For other recommendations:
> 
> 1. I'd prefer:
>    Introduce explicit CLI flags:
>    --std
>    --no-std
> 
> 3 and 4: LGTM. Please help to fix the 1. 3. 4. thanks.

It’s my pleasure — I’d be glad to work on that.
By the way, could you clarify whether we should use host or ca in this project?


> 5. The cargo-optee tool isn’t tied to the dev Docker image, so we can keep the current dev image unchanged and consider optimizing it later.
>    However, if we merge this PR, our documentation and CI will need a full update.
>    How about we add cargo-optee in this PR while keeping both the old Makefile system and the new tool? After this merged, we can update the docs and CI step by step, and remove the Makefiles once the new tool is stable and widely used. What do you think?

LGTM! I think this approach is great — it allows users to transition smoothly from the old Makefile workflow to the new tool.

DemesneGH:
> could you clarify whether we should use host or ca in this project?

@asir66 Let's use `host` which also aligns with optee C examples https://github.com/linaro-swg/optee_examples/tree/master/acipher

PR #255, opened 2025-11-17
examples: change for acipher-rs

ldshyu:
fix potential buffer overflow and match behavior with optee_examples

### DIFF
--- a/examples/acipher-rs/ta/src/main.rs
+++ b/examples/acipher-rs/ta/src/main.rs
@@ -97,8 +97,13 @@ fn encrypt(rsa: &mut RsaCipher, params: &mut Parameters) -> Result<()> {
             match cipher.encrypt(&[], plain_text) {
                 Err(e) => Err(e),
                 Ok(cipher_text) => {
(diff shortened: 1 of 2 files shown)


examples/acipher-rs/ta/src/main.rs
```diff
@@ -97,8 +97,13 @@ fn encrypt(rsa: &mut RsaCipher, params: &mut Parameters) -> Result<()> {
             match cipher.encrypt(&[], plain_text) {
                 Err(e) => Err(e),
                 Ok(cipher_text) => {
-                    p1.buffer().clone_from_slice(&cipher_text);
-                    Ok(())
+                    if cipher_text.len() > p1.buffer().len() {
+                        p1.set_updated_size(cipher_text.len());
+                        Err(optee_utee::Error::new(ErrorKind::ShortBuffer))
```
DemesneGH: Err(ErrorKind::ShortBuffer.into())

ldshyu: Thanks for reviewing.  I have updated the change accordingly.

PR #258, opened 2025-12-08
chore: get rid of now-stabilized error_in_core feature

TheButlah:
Attempting to build the SDK without stable compiler features is not possible right now due to the use of `#![feature(error_in_core)]`. However, this feature was [stabilized in rust 1.81](https://blog.rust-lang.org/2024/09/05/Rust-1.81.0/#what-s-in-1-81-0-stable) in September 2024.

It would be good to remove the now-unneeded unstable flag.

### DIFF
--- a/optee-utee/src/lib.rs
+++ b/optee-utee/src/lib.rs
@@ -16,7 +16,6 @@
 // under the License.
 
 #![cfg_attr(not(feature = "std"), no_std)]
-#![cfg_attr(not(feature = "std"), feature(error_in_core))]
 
 // Requires `alloc`.
 #[macro_use]


ivila:
@DemesneGH Should we update our patched version? We’re starting to run into more issues as it ages — the current patch is from 2024-05-14.

TheButlah:
in the meantime then, ill update my PR with a path forward:
* Ill detect the rustc version being used in the build script
* If its too old, ill enable the nightly feature, otherwise I wont.
* It would continue to support the older MSRV

This would unblock using stable toolchains for no-std users, without requiring yall to upgrade the repo's rust-toolchain.toml

PR #260, opened 2025-12-09
chore: get rid of now-stabilized error_in_core feature

TheButlah:
(continuation of #258 )

Attempting to build the SDK without stable compiler features is not possible right now due to the use of `#![feature(error_in_core)]`. However, this feature was [stabilized in rust 1.81](https://blog.rust-lang.org/2024/09/05/Rust-1.81.0/#what-s-in-1-81-0-stable) in September 2024.

It would be good to remove the now-unneeded unstable flag.

### DIFF
--- a/optee-utee/build.rs
+++ b/optee-utee/build.rs
@@ -1,19 +1,34 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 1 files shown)


optee-utee/build.rs
```diff
@@ -0,0 +1,27 @@
...
+/// Outputs unstable #[feature = "foo"] iff the rustc version is older than $version
+macro_rules! maybe_feat {
+    ($out:expr, $feat:literal, $version:literal) => {
+        {
+            let filename = $out.join(concat!($feat, ".rs"));
+            
+            let s = if version_check::is_min_version($version).unwrap_or(false) {
+                String::new()
+            } else {
+                format!("#![feature = \"{}\"]\n", $feat)
```
ivila: Should be: `format!("#![feature({})]\n", $feat)`

optee-utee/build.rs
```diff
@@ -0,0 +1,27 @@
+use std::fs;
```
ivila, line 18: Please add a valid license header.

ivila:
Fixed by #262

PR #261, opened 2025-12-09
feat: support target-specific specialization of OPTEE_CLIENT_EXPORT

### LINKED ISSUE #259: Support target-specific OPTEE_CLIENT_EXPORT env variables to simplify cross compilation
# Proposed Change:
Introduce target-specific variants `OPTEE_CLIENT_EXPORT` for x86 and aarch64, using same style as [pkg-config](https://docs.rs/pkg-config/0.3.32/pkg_config/#cross-compilation). This allows different cross compilation targets to have their env variables simultaneously present.

# Details
We add variants of `OPTEE_CLIENT_EXPORT` that can be specialized by their target architecture via a suffix, in the following prioritized order:

1. `OPTEE_CLIENT_EXPORT_<target>` - for example, `OPTEE_CLIENT_EXPORT_x86_64-unknown-linux-gnu`
2. `OPTEE_CLIENT_EXPORT_<target_with_underscores>` - for example, `OPTEE_CLIENT_EXPORT_x86_64_unknown_linux_gnu`
3. `OPTEE_CLIENT_EXPORT` - same as it is today

> NOTE: `<target>` is determined in the build script via the [environment variables cargo sets for build scripts](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts)

# How is it different from what we have today?
Today, teaclave CAs can already be compiled for x86 - simply build optee_client for x86, and provide that as your `OPTEE_CLIENT_EXPORT` variable. However, if you want to be able to target *both* x86 or aarch64, it implies that you have to switch this environment variable every time before you build. This becomes especially complicated in some develoment environments - for example, when using the [optee-client packaged by nix](https://search.nixos.org/packages?channel=25.11&show=optee-client), the conventional way would be to create an nix dev shell that sets an environment variable to the evaluated package, and it would be quite annoying to have to switch between an x86 and an aarch64 dev shell every time I want to change the target architecture.

Instead, this proposal allows *both* environment variables to be provided simultaneously, and allows the build script to choose which one to use based on the `--target` passed to `cargo build`.

# Why even care about CAs targetting x86 in the first place?
I ackowledge that CAs that target x86 don't intrinsically make much sense (because TAs inherently must be aarch64, and CAs are only useful with corresponding TAs), but the optee_client C code *does* compile for x86 and it would add fewer caveats around the build system if it can target either x86 or aarch64. 

Reducing friction for integrating teaclave into existing coebases is important for adoption. A concrete example of why this feature would reduce tooling pain is that at work, the monorepo we use assumes that *all* crates in the cargo workspace can target *both* aarch64 and x86 linux. TAs intrinsically cannot, so that implies there is at least two cargo workspaces (one for non-TA code and one for TAs). If CAs cannot target x86, then the number of workspaces increases to 3 (non-optee, CAs, and TAs). This could be simplified by enabling CAs to target x86.

Furthermore, LSPs typically assume that they are targetting the host architecture, and for most devices thats x86. While its possible to configure this, many devs don't even know how, so they would just have a degraded development experience.

# PR status
https://github.com/apache/teaclave-trustzone-sdk/pull/261

TheButlah:
Closes #259

### DIFF
--- a/optee-teec/optee-teec-sys/build.rs
+++ b/optee-teec/optee-teec-sys/build.rs
@@ -18,6 +18,47 @@
 use std::env::{self, VarError};
 use std::path::PathBuf;
 
+const ENV_OPTEE_CLIENT_EXPORT: &str = "OPTEE_CLIENT_EXPORT";
+const AARCH64: &str = "aarch64-unknown-linux-gnu";
+const X86_64: &str = "x86_64-unknown-linux-gnu";
+
+#[derive(Debug, Clone, Copy)]
+enum Target {
+    Aarch64,
+    X86_64,
+}
+
+impl Target {
+    fn get() -> Result<Target, String> {
+        let var = env::var("TARGET").expect("infallible");
+        match var.as_str() {
+            AARCH64 => Ok(Target::Aarch64),
+            X86_64 => Ok(Target::X86_64),
+            _ => Err(var),
+        }
+    }
+}
+
+/// Lists out env vars to check in order of priority
+fn enumerate_vars_for_target() -> Vec<String> {
+    let prefix = ENV_OPTEE_CLIENT_EXPORT.to_owned();
+    let Ok(target) = Target::get() else {
+        // Unknown targets should not try using the suffixed variants.
+        return vec![prefix];
+    };
+    let aarch64 = format!("_{AARCH64}");
+    let x86 = format!("_{X86_64}");
+    let suffixes = match target {
+        Target::Aarch64 => &[&aarch64, &aarch64.replace("-", "_"), ""],
+        Target::X86_64 => &[&x86, &x86.replace("-", "_"), ""],
+    };
+
+    suffixes
+        .iter()
+        .map(|s| format!("{prefix}{s}"))
+        .collect()
+}
+
 fn main() -> Result<(), VarError> {
     if !is_feature_enable("no_link")? {
         link();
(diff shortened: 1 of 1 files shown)


optee-teec/optee-teec-sys/build.rs
```diff
@@ -18,6 +18,47 @@
 use std::env::{self, VarError};
 use std::path::PathBuf;
 
+const ENV_OPTEE_CLIENT_EXPORT: &str = "OPTEE_CLIENT_EXPORT";
+const AARCH64: &str = "aarch64-unknown-linux-gnu";
```
ivila, line 22: The targets aren’t limited to `aarch64-unknown-linux-gnu` and `x86_64-unknown-linux-gnu`. They can also include `arm-unknown-linux-gnueabihf` (see: [arm-unknown-optee.json](https://github.com/apache/teaclave-trustzone-sdk/blob/main/arm-unknown-optee.json)) and other variants, such as `aarch64-unknown-none` for `no_std` tests. It would be better to detect the architecture from the target triple rather than relying on the entire triple string.

ivila: Suggest using [CARGO_CFG_TARGET_ARCH](https://doc.rust-lang.org/cargo/reference/environment-variables.html) to significantly reduce the effort required here.

PR #263, opened 2025-12-15
feat: add cargo-optee and update the docs of quick start

asir66:
# Overview

This update is based on PR #245 and focuses on improving the robustness of cargo-optee in several areas, including manifest path resolution, std/no-std argument semantics, and consistency across documentation and CI.

Specifically, this PR includes the following changes:

1. Fix issues in cargo-optee related to manifest path resolution, including workspace scenarios
2. Clarify and improve the semantics and parsing logic of `--std` / `--no-std`
3. Normalize JSON file field ordering
4. Update the Quick Start documentation and CI configuration
5. Summarize differences between cargo, cargo-sgx, and cargo-optee based on investigation, and outline potential future directions

Related background discussion: https://github.com/apache/teaclave-trustzone-sdk/pull/245#issuecomment-3584945539

# 1. Manifest Path Resolution Fix

This issue is fundamentally caused by insufficiently robust manifest path resolution. It manifests differently depending on the project structure:

1. Single-package projects  
When `--manifest-path Cargo.toml` (without `./`) is explicitly provided, cargo-optee does not correctly handle this edge case when deriving the project path, which may result in failing to locate the project directory.

2. Workspace projects  
When a relative path containing `./` is used as the manifest path, cargo metadata returns canonicalized manifest paths, while cargo-optee attempts to match them using the original, non-normalized path string. This mismatch causes package discovery to fail.

## Fix and Rationale

To address this issue at its root, manifest paths are now canonicalized at the entry point using:
```rust
std::fs::canonicalize()
```

This standard library function normalizes paths into absolute form and automatically eliminates:
- `./`
- `../`

# 2. `--std` / `--no-std` Argument Semantics

## Previous Behavior

cargo-optee previously used a boolean flag to represent whether std was enabled.
When the flag was not explicitly provided, it defaulted to Some(false), which
caused cargo-optee to **always fall back to the no-std build mode.**

## Updated Design

This update introduces an explicit `--no-std` flag and refactors the argument
parsing logic accordingly:

- `--std` and `--no-std` are mutually exclusive
- If neither is specified, the internal state is set to `None`
- The final build mode is determined according to the defined precedence rules

This clarifies user intent and improves extensibility for future configuration
sources.

# 3. JSON File Normalization

Relevant JSON files have been updated to ensure that fields are ordered in
ascending ASCII order.
- This change does not affect functionality, but improves:
- file consistency
- diff readability

alignment with common tooling and formatting conventions

# 4. Documentation and CI Updates

- The Quick Start guide has been updated to recommend the cargo-optee workflow as the primary approach
- Documentation sections that were inconsistent with current behavior have been corrected
- The CI configuration has been updated to cover the new arguments and corrected behavior

During this process, an additional documentation inconsistency was identified and addressed as part of this PR.

# 5. Investigation: Differences Between cargo, cargo-sgx, and cargo-optee

The following section summarizes findings from an investigation into the design and usage differences between cargo, cargo-sgx, and cargo-optee.

> The content in this section is exploratory and design-oriented, and does not affect the functional fixes or merge decision of this PR.

## 5.1 cargo-optee Is Not a cargo Subcommand

Currently, cargo-optee cannot be invoked as a cargo subcommand (i.e. cargo optee).

One possible reason is that the customized cargo toolchain used in the environment does not include subcommand extension support.

From a pragmatic perspective, the current form is acceptable. However, from a consistency and extensibility standpoint, retaining this capability may be worth considering in the future.

## 5.2 Rationale for `cargo-optee build ta/ca`

Unlike cargo-sgx, which does not require explicit sub-target selection, cargo-optee requires specifying either `ta` or `ca`.

This difference stems from fundamental architectural distinctions:
- In Intel SGX, enclaves are closer to being linked as libraries
- In TrustZone, Trusted Applications are standalone secure-world entities

As a result, this design choice is both necessary and appropriate.

## 5.3 Default Build Mode Differences

<html><head></head><body>
Tool | Default Build Mode
-- | --
cargo | `--debug`
cargo-sgx | `--release`
cargo-optee | `--release`

</body></html>

cargo-optee aligns with cargo-sgx in this respect, which matches the performance and determinism requirements of TEE environments.

## 5.4 `run` vs `sync`

Both cargo and cargo-sgx support running applications directly via `run`. In contrast, cargo-optee requires an explicit `sync` step before execution.

This difference is driven by runtime environment constraints:
- SGX relies on real hardware and encapsulates the toolchain tightly
- TrustZone requires explicit deployment of TA and CA artifacts to either an emulator or a physical device

Therefore, this behavior reflects architectural constraints rather than a tooling deficiency.

That said, it may be worth exploring whether the `sync` step could be optionally integrated into cargo-optee (e.g. via a flag or default behavior) to improve usability.

## 5.5 Features, Targets, and Workflow Overlap

In cargo:
- features control code paths
- targets select toolchains

In cargo-optee, these dimensions partially overlap at the workflow level.

This has also been a key point in previous discussions(see: https://github.com/apache/teaclave-trustzone-sdk/pull/245#discussion_r2462214183).

At present, the main source of complexity comes from the overlap between features and targets in the build workflow. This overlap makes certain parts of the model harder to reason about and evolve.

It is possible that this issue could be addressed more cleanly in the future if the toolchain converges on a cargo-only workflow (for example, if xargo is eventually phased out). Under such conditions, the separation of concerns between features and targets may become clearer.

For now, however, cargo-optee remains a well-suited tool for this project: it effectively reduces configuration burden, integrates smoothly with the TrustZone SDK, and provides a simple and efficient developer experience.

## 6. Future Directions (Design Considerations)

In the long term, cargo-optee could consider adopting a unified and configurable parameter resolution model, for example:
```
CLI arguments > Cargo.toml > environment variables > defaults
```

(Environment variables may also be viewed as overridable defaults.)

Such a model could:
- reduce the need for verbose command-line invocations
- improve build reproducibility
- better align cargo-optee with Cargo’s configuration philosophy

This would allow for unified configurability without sacrificing explicitness.

Thanks the help of @DemesneGH @m4sterchain

### DIFF
--- a/cargo-optee/README.md
+++ b/cargo-optee/README.md
@@ -1,10 +1,14 @@
 # cargo-optee
 
-A Cargo subcommand for building OP-TEE Trusted Applications (TAs) and Client Applications (CAs) in Rust.
+A Cargo subcommand for building OP-TEE Trusted Applications (TAs) and Client
+Applications (CAs) in Rust.
 
 ## Overview
 
-`cargo-optee` simplifies the development workflow for OP-TEE applications by replacing complex Makefiles with a unified, type-safe command-line interface. It handles cross-compilation, custom target specifications, environment setup, and signing automatically.
+`cargo-optee` simplifies the development workflow for OP-TEE applications by
+replacing complex Makefiles with a unified, type-safe command-line interface. It
+handles cross-compilation, custom target specifications, environment setup, and
+signing automatically.
 
 ## High-Level Design
 
@@ -59,83 +63,124 @@ A Cargo subcommand for building OP-TEE Trusted Applications (TAs) and Client App
 
 ### Installation
 
-Assume developers have Rust, Cargo, and the gcc toolchain installed and added to PATH (the guide is in future plan). Then install `cargo-optee` using Cargo:
+Assume developers have Rust, Cargo, and the gcc toolchain installed and added to
+PATH (the guide is in future plan). Then install `cargo-optee` using Cargo:
 
 ```bash
 cargo install cargo-optee
 ```
 
+### Quick Build for Hello World
+
+This section provides a quick start guide for building the Hello World example
+using `cargo-optee`. Before proceeding, ensure you have set up the Docker
+development environment. For detailed instructions on setting up the Docker
+environment, refer to the [QEMU emulation
+guide](../docs/emulate-and-dev-in-docker.md).
+
+#### Prerequisites
+
+First, pull and run the Docker image. We provide two types of images:
+- **no-std environment**: For building TAs without the standard library
+  (recommended for quick start)
+- **std environment**: For building TAs with Rust standard library support
+
+**Pull and start the no-std development environment:**
+```bash
+# Pull the pre-built development environment image
+docker pull teaclave/teaclave-trustzone-emulator-nostd-expand-memory:latest
+
+# Start the development container
+docker run -it --rm \
+  --name teaclave_dev_env \
+  -v $(pwd):/root/teaclave_sdk_src \
+  -w /root/teaclave_sdk_src \
+  teaclave/teaclave-trustzone-emulator-nostd-expand-memory:latest
+```
+
+> 📖 **Note**: If you need Rust standard library (std) support, refer to the
+> [Docker development environment guide with std
+> support](../docs/emulate-and-dev-in-docker-std.md) and use the
+> `teaclave/teaclave-trustzone-emulator-std-expand-memory:latest` image.
+
+#### Build Steps
+
+**1. Navigate to the Hello World example directory**
+
+Inside the Docker container, execute:
+```bash
+cd examples/hello_world-rs/
+```
+
+**2. Build the Trusted Application (TA)**
+
+Build the TA using `cargo-optee`. In the Docker environment, the OP-TEE TA
+development kit is typically located at
+`/opt/teaclave/optee/optee_os/out/arm-plat-vexpress/export-ta_arm64`:
+
+```bash
+# Build aarch64 no-std TA (default configuration)
+cargo-optee build ta \
+  --manifest-path ta/Cargo.toml \
+  --ta-dev-kit-dir /opt/teaclave/optee/optee_os/out/arm-plat-vexpress/export-ta_arm64 \
+  --arch aarch64 \
+  --no-std
+```
+
+**3. Build the Client Application (CA)**
+
+Build the client application:
+```bash
+# Build aarch64 CA
+cargo-optee build ca \
+  --manifest-path host/Cargo.toml \
+  --optee-client-export /opt/teaclave/optee/optee_client/export_arm64 \
+  --arch aarch64
+```
+
+> 💡 **Tip**: If you configure metadata in your `Cargo.toml` file (see
+> [Configuration System](#configuration-system)), you can simplify the command
+> by only specifying `--manifest-path`:
+> ```bash
+> cargo-optee build ca --manifest-path host/Cargo.toml
+> ```
+
+#### Build Output
+
+After a successful build, you can find the generated files at the following
+locations:
+
+- **TA binary**:
+  `ta/target/aarch64-unknown-linux-gnu/release/133af0ca-bdab-11eb-9130-43bf7873bf67.ta`
+- **CA binary**: `host/target/aarch64-unknown-linux-gnu/release/hello_world-rs`
+
+#### Next Steps
+
+After building, you can:
+1. Use the `sync_to_emulator` command to sync build artifacts to the emulator
+   environment
+2. Start the QEMU emulator for testing
+3. Refer to the [QEMU emulation guide](../docs/emulate-and-dev-in-docker.md) for
+   complete development and testing workflows
+
+> 💡 **Tip**: If you configure metadata in `Cargo.toml` (see [Configuration
+> System](#configuration-system)), you can simplify build commands by only
+> specifying `--manifest-path`.
+
 ## Configuration System
 
-`cargo-optee` uses a flexible configuration system with the following priority (highest to lowest):
+`cargo-optee` uses a flexible configuration system with the following priority
+(highest to lowest):
 
-1. **Command Line Arguments** - Direct CLI flags override everything
-2. **Cargo.toml Metadata** - Project-specific configuration in `[package.metadata.optee.*]` sections  
+1. **Command Line Arguments** - Direct CLI flags override everything (see [Build
+   through CLI](#build-through-cli))
+2. **Cargo.toml Metadata** - Project-specific configuration in
+   `[package.metadata.optee.*]` sections (see [Build through
+   metadata](#build-through-metadata))
 3. **Defaults** - Built-in sensible defaults
 
-This allows projects to define their standard configuration in `Cargo.toml` while still permitting CLI overrides for specific builds.
-
-### Metadata Configuration
-
-#### Trusted Application (TA) Metadata
-
-Configure TA builds in your `Cargo.toml`:
-
-```toml
-[package.metadata.optee.ta]
-arch = "aarch64"                    # Target architecture: "aarch64" | "arm" (optional, default: "aarch64")
-debug = false                       # Debug build: true | false (optional, default: false)
-std = false                         # Use std library: true | false (optional, default: false)
-uuid-path = "../uuid.txt"           # Path to UUID file (optional, default: "../uuid.txt")
-# Architecture-specific configuration (omitted architectures default to null/unsupported)
-ta-dev-kit-dir = { aarch64 = "/opt/optee/export-ta_arm64", arm = "/opt/optee/export-ta_arm32" }
-signing-key = "/path/to/key.pem"    # Path to signing key (optional, defaults to ta-dev-kit/keys/default_ta.pem)
-```
-
-**Allowed entries:**
-- `arch`: Target architecture (`"aarch64"` or `"arm"`)
-- `debug`: Build in debug mode (`true` or `false`) 
-- `std`: Enable std library support (`true` or `false`)
-- `uuid-path`: Relative or absolute path to UUID file
-- `ta-dev-kit-dir`: Architecture-specific paths to TA development kit (required)
-- `signing-key`: Path to signing key file
-
-#### Client Application (CA) Metadata
-
-Configure CA builds in your `Cargo.toml`:
-
-```toml
-[package.metadata.optee.ca]
-arch = "aarch64"                    # Target architecture: "aarch64" | "arm" (optional, default: "aarch64")
-debug = false                       # Debug build: true | false (optional, default: false)
-# Architecture-specific configuration
-# if your CA only supports aarch64, you can omit arm
-optee-client-export = { aarch64 = "/opt/optee/export-client_arm64" } 
-```
-
-**Allowed entries:**
-- `arch`: Target architecture (`"aarch64"` or `"arm"`)
-- `debug`: Build in debug mode (`true` or `false`)
-- `optee-client-export`: Architecture-specific paths to OP-TEE client export (required)
-
-#### Plugin Metadata
-
-Configure plugin builds in your `Cargo.toml`:
-
-```toml
-[package.metadata.optee.plugin]
-arch = "aarch64"                    # Target architecture: "aarch64" | "arm" (optional, default: "aarch64")  
-debug = false                       # Debug build: true | false (optional, default: false)
-uuid-path = "../plugin_uuid.txt"    # Path to UUID file (required for plugins)
-# Architecture-specific configuration
-optee-client-export = { aarch64 = "/opt/optee/export-client_arm64", arm = "/opt/optee/export-client_arm32" }
-```
-
-**Allowed entries:**
-- `arch`: Target architecture (`"aarch64"` or `"arm"`)
-- `debug`: Build in debug mode (`true` or `false`)
-- `uuid-path`: Relative or absolute path to UUID file (required for plugins)
-- `optee-client-export`: Architecture-specific paths to OP-TEE client export (required)
+This allows projects to define their standard configuration in `Cargo.toml`
+while still permitting CLI overrides for specific builds.
 
 ### Project Structure
 
@@ -159,10 +204,85 @@ project/
         └── lib.rs
 ```
 
-See examples in the SDK for reference, such as `hello_world-rs`.
-The `cargo new` command (planned, not yet available) will generate a project template with this structure. For now, copy an existing example as a starting point.
+See examples in the SDK for reference, such as `hello_world-rs`. The `cargo new`
+command (planned, not yet available) will generate a project template with this
+structure. For now, copy an existing example as a starting point.
 
-### Build Commands
+### Usage Workflows (including future design)
+
+#### Development/Emulation Environment
+
+For development and emulation, developers would like to build the one project
+and deploy to a target filesystem (e.g. QEMU shared folder) quickly. Frequent
+builds and quick rebuilds are common.
+
+**Using CLI arguments:**
+```bash
+# 1. Create new project (future)
+cargo-optee new my_app
+cd my_app
+
+# 2. Build TA and CA
+cargo-optee build ta \
+  --ta-dev-kit-dir $TA_DEV_KIT_DIR \
+  --manifest-path ./ta/Cargo.toml \
+  --arch aarch64 \
+  --std
+
+cargo-optee build ca \
+  --optee-client-export $OPTEE_CLIENT_EXPORT \
+  --manifest-path ./host/Cargo.toml \
+  --arch aarch64
+
+# 3. Install to specific folder (future), e.g. QEMU shared folder for emulation
+cargo-optee install --target /tmp/qemu-shared-folder
+```
+
+**Using metadata configuration:**
+```bash
+# 1. Configure once in Cargo.toml files, then simple builds
+cd ta && cargo-optee build ta
+cd ../host && cargo-optee build ca
+
+# 2. Override specific parameters when needed
+cd ta && cargo-optee build ta --debug  # Override to debug build
+cd host && cargo-optee build ca --arch arm  # Override architecture
+```
+
+#### Production/CI Environment
+
+For production and CI environments, artifacts should be cleaned up after
+successful builds. It can help to avoid storage issues on CI runners.
+
+**Automated Build Pipeline:**
+```bash
+#!/bin/bash
+# CI build script
+
+set -e
+
+# Build TA (release mode)
+cargo-optee build ta \
+  --ta-dev-kit-dir $TA_DEV_KIT_DIR \
+  --manifest-path ./ta/Cargo.toml \
+  --arch aarch64 \
+  --std \
+  --signing-key ./keys/production.pem
+
+# Build CA (release mode)
+cargo-optee build ca \
+  --optee-client-export $OPTEE_CLIENT_EXPORT \
+  --manifest-path ./host/Cargo.toml \
+  --arch aarch64
+
+# Install to staging area (future)
+cargo-optee install --target ./dist
+
+# Clean build artifacts to save space (future)
+cargo-optee clean --all
+```
+
+### Build through CLI
 
 #### Build Trusted Application (TA)
 
@@ -179,7 +299,8 @@ cargo-optee build ta \
 ```
 
 **Required:**
-- `--ta-dev-kit-dir <PATH>`: Path to OP-TEE TA development kit (available after building OP-TEE OS), user must provide this for building TAs.
+- `--ta-dev-kit-dir <PATH>`: Path to OP-TEE TA development kit (available after
+  building OP-TEE OS), user must provide this for building TAs.
 
 **Optional:**
 - `--manifest-path <PATH>`: Path to Cargo.toml manifest file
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -35,6 +35,8 @@ The Docker image automates the entire setup process for TrustZone emulation
 in QEMU, enabling you to focus on writing and testing your applications 
 efficiently, without the hassle of manual configuration.
 
+**UPDATES** 我们更新了一个更加方便的工具，更好的封装让开发者更加便捷[cargo-optee](./cargo-optee/README.md), 后续我们可能会慢慢放弃makefile结合cargo的方式，使用更加rust的方式开发. 为了平稳过度，我们仍然保留的使用Makefile进行构建的相关工具组件。
```
DemesneGH: Could you provide an English version to keep consistency in this doc? Thanks

docs/emulate-and-dev-in-docker-std.md
```diff
@@ -19,9 +29,16 @@ The **dev-env with std support** enables **developing TA using Rust std** by pro
 - **Rust standard library tailored for OP-TEE** - Build TAs using collections, networking, serialization capabilities
 - **Mixed development support** - Combine different host and TA architectures, including switching between no-std/std in the same project
 
-## 1. Setting Up the Dev-Env with Std Support
+## Workflow: Developing std-enabled TAs with cargo-optee
```
DemesneGH: Recommend to reorganize:
```
## Setting Up the Dev-Env with Std Support
...
## Build STD TA
### 1. cargo-optee ...
### 2. make ...
```

This doc just provides the command to build STD TAs, for the description of cargo-optee, we can add a link to the cargo-optee README. For the no-std TAs, we can provides the building command in the other emulate-... doc.

Please also reorganize the another emulate-... doc, the goal is to make each doc clear and avoid duplicated contents.

DemesneGH:
Please also fix the CI, thanks!

asir66:
> Please also fix the CI, thanks!请同时修复 CI，谢谢！

Sorry, it's my fault. I will fix the CI and doc. thank you.

cargo-optee/src/ca_builder.rs
```diff
@@ -0,0 +1,281 @@
...
+    let target_dir = workspace_target_dir.join(target).join(profile);
+
+    // Get the library name from Cargo.toml
+    let cargo_toml = std::fs::read_to_string("Cargo.toml")?;
+    let lib_name = cargo_toml
+        .lines()
+        .find(|line| line.trim().starts_with("name"))
+        .and_then(|line| line.split('=').nth(1))
+        .map(|s| s.trim().trim_matches('"'))
+        .ok_or_else(|| anyhow::anyhow!("Could not find package name in Cargo.toml"))?;
```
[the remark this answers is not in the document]
DemesneGH: Please help to fix this and the next two Copilot comments @asir66

cargo-optee/src/main.rs
```diff
@@ -0,0 +1,567 @@
...
+    let ta_dev_kit_dir_config = build_config.require_ta_dev_kit_dir()?;
+    let ta_dev_kit_dir = resolve_path_relative_to_project(&ta_dev_kit_dir_config, &project_path);
+
+    // Validate that ta_dev_kit_dir exists (print absolute path)
+    if !ta_dev_kit_dir.exists() {
+        bail!(
+            "TA development kit directory does not exist: {:?}",
+            ta_dev_kit_dir
+        );
+    }
```
DemesneGH: Move this check into resolve_path_relative_to_project() to reduce the duplicate code, the same case is in line 115-118, please also check and fix other similar cases

cargo-optee/src/main.rs
```diff
@@ -0,0 +1,567 @@
...
+    features: Option<String>,
+}
+
+#[derive(Debug, Parser)]
+#[clap(version = env!("CARGO_PKG_VERSION"))]
+#[clap(about = "Build tool for OP-TEE Rust projects")]
+pub(crate) struct Cli {
+    #[clap(subcommand)]
+    cmd: Command,
+}
```
DemesneGH: Recommend to move all cli structs definition to another file `cli.rs` to make the `main.rs` simple and clean

cargo-optee/src/config.rs
```diff
@@ -0,0 +1,637 @@
...
+
+impl BuildConfig {
+    /// Create a new build config by resolving parameters with priority:
+    /// 1. Command line arguments (highest priority)
+    /// 2. [package.metadata.optee.<component_type>] in Cargo.toml
+    /// 3. Default values or error for mandatory parameters
+    #[allow(clippy::too_many_arguments)]
+    pub fn resolve(
+        project_path: &Path,
+        component_type: &str, // "ta", "ca", or "plugin"
```
DemesneGH: component_type: use the enum, not string

cargo-optee/src/main.rs
```diff
@@ -0,0 +1,567 @@
...
+        } else {
+            // Canonicalize must succeed, otherwise treat as invalid manifest path
+            parent
+                .canonicalize()
+                .map_err(|_| anyhow::anyhow!("Invalid manifest path"))?
+        }
+    } else {
+        std::env::current_dir()?
+    };
+
```
DemesneGH, line 157: Move this into a fn for code reuse, same code in line158-175

cargo-optee/src/main.rs
```diff
@@ -0,0 +1,567 @@
...
+        Command::Build(build_cmd) => match build_cmd {
+            BuildCommand::TA { build_cmd } => {
+                // Convert bool flags to Option<bool>: --std -> Some(true), --no-std -> Some(false), neither -> None
+                let std = if build_cmd.std {
+                    Some(true)
+                } else if build_cmd.no_std {
+                    Some(false)
+                } else {
+                    None
+                };
```
DemesneGH, line 69: Move this into a fn to reduce code, try to simplify like:
```
        match (build_cmd.std, build_cmd.no_std) {
            (true, false) => Some(true),
            (false, true) => Some(false),
            _ => None,
        }
```

cargo-optee/src/main.rs
```diff
@@ -0,0 +1,567 @@
...
+        Command::Clean { clean_cmd } => {
+            let project_path = if let Some(manifest) = clean_cmd.manifest_path {
+                manifest
+                    .parent()
+                    .ok_or_else(|| anyhow::anyhow!("Invalid manifest path"))?
+                    .to_path_buf()
+            } else {
+                std::env::current_dir()?
+            };
+
```
DemesneGH: Also can reuse some fn like `resolve_project_path()`

cargo-optee/src/ta_builder.rs
```diff
@@ -0,0 +1,505 @@
...
+        .and_then(|v| v.as_str())
+        .ok_or_else(|| anyhow::anyhow!("Could not get target directory from cargo metadata"))?;
+
+    let target_dir = PathBuf::from(target_directory);
+    let profile_dir = target_dir.join(target).join(profile);
+    let binary_path = profile_dir.join(&package_name);
+
+    if !binary_path.exists() {
+        bail!("Binary not found at {:?}", binary_path);
+    }
```
DemesneGH: All the same pattern can move into a fn like `join_and_check()`, using that fn avoids we forget  check the path

cargo-optee/src/main.rs
```diff
@@ -0,0 +1,567 @@
...
+    if let Some(eq_pos) = s.find('=') {
+        let (key, value) = s.split_at(eq_pos);
+        let value = &value[1..]; // Skip the '=' character
+        Ok((key.to_string(), value.to_string()))
+    } else {
+        Err(format!(
+            "Invalid environment variable format: '{}'. Expected 'KEY=VALUE'",
+            s
+        ))
+    }
```
DemesneGH: Try to use this:
```
    s.split_once('=')
        .map(|(key, value)| (key.to_string(), value.to_string()))
        .ok_or_else(|| format!(
            "Invalid environment variable format: '{}'. Expected 'KEY=VALUE'",
            s
        ))
```

cargo-optee/src/main.rs
```diff
@@ -0,0 +1,567 @@
...
+}
+
+/// Setup cargo environment by checking availability and sourcing environment if needed
+fn setup_cargo_environment() -> anyhow::Result<()> {
+    // Check if cargo is available
+    let cargo_available = std::process::Command::new("which")
+        .arg("cargo")
+        .output()
+        .map(|output| output.status.success())
+        .unwrap_or(false);
```
DemesneGH: Try this:
```
    let cargo_available = std::process::Command::new("which")
        .arg("cargo")
        .output()
        .is_ok_and(|output| output.status.success());
```

ivila:
For **5.1, "cargo-optee" is not recognized as a cargo subcommand**, it is because cargo invokes external binaries in an unexpected way.

For example, `cargo optee build` is invoked as `cargo-optee optee build`. This extra "optee" argument causes the subcommand parser to fail.

One workaround is to add a parsing rule or configuration option that discards the first (duplicate) argument:
```rust
#[derive(Parser)]
#[command(version, about, long_about = None)]
struct Cli {
    // By stripping the duplicate "optee" argument during subcommand invocation, 
    // this solution maintains compatibility between standalone and cargo-integrated execution modes.
    #[arg(name = "tool_name", hide = true)]
    _tool_name: String,
    #[command(subcommand)]
    commands: Commands,
}
```

asir66:
> For **5.1, "cargo-optee" is not recognized as a cargo subcommand**, it is because cargo invokes external binaries in an unexpected way.
> 
> For example, `cargo optee build` is invoked as `cargo-optee optee build`. This extra "optee" argument causes the subcommand parser to fail.
> 
> One workaround is to add a parsing rule or configuration option that discards the first (duplicate) argument:
> 
> ```rust
> #[derive(Parser)]
> #[command(version, about, long_about = None)]
> struct Cli {
>     // By stripping the duplicate "optee" argument during subcommand invocation, 
>     // this solution maintains compatibility between standalone and cargo-integrated execution modes.
>     #[arg(name = "tool_name", hide = true)]
>     _tool_name: String,
>     #[command(subcommand)]
>     commands: Commands,
> }
> ```

That makes sense—thank you for the suggestion.

I also noticed that with the previous implementation, running cargo-optee in standalone mode results in the following error:
```
error: the following required arguments were not provided:
  <tool_name>
```

For this reason, I believe the following approach is more robust: making _tool_name an optional argument.
```rust
#[derive(Debug, Parser)]
#[command(version, about)]
pub struct Cli {
    // By stripping the duplicate "optee" argument during subcommand invocation,
    // this solution maintains compatibility between standalone and cargo-integrated execution modes.
    // When invoked as `cargo optee build`, cargo passes "optee" as the first argument.
    // When invoked as `cargo-optee build`, no extra argument is passed.
    #[arg(name = "tool_name", hide = true)]
    _tool_name: Option<String>,

    #[command(subcommand)]
    pub cmd: Command,
}
```

This allows the CLI to work correctly in both invocation scenarios without requiring additional conditional logic.

DemesneGH:
@asir66 Thanks for your effort!
I've checked the implementation in `cargo-sgx`:
```
#[derive(Debug, Parser)]
#[clap(bin_name = "cargo")]
#[clap(version = VERSION)]
pub(crate) enum Opts {
    /// Utilities to build sgx program.
    #[clap(name = "sgx")]
    #[clap(version = VERSION)]
    #[clap(action = ArgAction::DeriveDisplayOrder)]
    Sgx(SgxArgs),
}

#[derive(Debug, clap::Args)]
pub(crate) struct SgxArgs {
    #[clap(subcommand)]
    cmd: Command,
}

#[derive(Debug, clap::Subcommand)]
enum Command {
    #[clap(name = "build")]
    Build(BuildCommand),
    #[clap(name = "run")]
    Run(RunCommand),
...
}
```
It seems `cargo-sgx` only supports being invoked as cargo-sgx sgx xxx and doesn't support the two scenarios.

Could you help with a brief study of how other cargo subcommands handle this case?

To achieve the goal of "cargo-optee acting as a cargo subcommand," it's good to follow the established principles used by other subcommands. If there is a standard convention, we can implement it in the same way; otherwise, the current implementation is fine.

examples/hello_world-rs/ta/Cargo.toml
```diff
@@ -40,3 +40,11 @@ optee-utee-build = { path = "../../../optee-utee-build" }
...
 lto = true
 opt-level = 1
+
+[package.metadata.optee.ta]
+arch = "aarch64"
+debug = false
+std = false
+uuid-path = "../uuid.txt"
+ta-dev-kit-dir = { aarch64 = "/opt/teaclave/optee/optee_os/out/arm-plat-vexpress/export-ta_arm64", arm = "/opt/teaclave/optee/optee_os/out/arm-plat-vexpress/export-ta_arm32" }
+signing-key = "/opt/teaclave/optee/optee_os/out/arm-plat-vexpress/export-ta_arm64/keys/default_ta.pem"
```
DemesneGH: These paths can be kept for demonstration, but should be removed before merging.

cargo-optee/src/main.rs
```diff
@@ -0,0 +1,402 @@
...
+    /// Expects a directory
+    Directory,
+    /// Expects a file
+    File,
+}
+
+fn main() {
+    env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info"))
+        .format_timestamp_millis()
+        .init();
```
DemesneGH: Seems we don't need the logger, we use print for output. Recommend removing these lines.

cargo-optee/src/main.rs
```diff
@@ -0,0 +1,402 @@
...
+                build_cmd,
+            } => {
+                // Convert bool flags to Option<bool>: --std -> Some(true), --no-std -> Some(false), neither -> None
+                let std = if build_cmd.std {
+                    Some(true)
+                } else if build_cmd.no_std {
+                    Some(false)
+                } else {
+                    None
+                };
```
DemesneGH: Same as line79?

examples/tls_client-rs/ta/Cargo.toml
```diff
@@ -45,3 +49,11 @@ optee-utee-build = { path = "../../../optee-utee-build" }
 panic = "abort"
 lto = false
 opt-level = 3
+
+[package.metadata.optee.ta]
+arch = "aarch64"
+debug = false
+std = true
+uuid-path = "../uuid.txt"
+ta-dev-kit-dir = { aarch64 = "/opt/teaclave/optee/optee_os/out/arm-plat-vexpress/export-ta_arm64", arm = "/opt/teaclave/optee/optee_os/out/arm-plat-vexpress/export-ta_arm32" }
```
ivila: Do not hardcode them here, as doing so may cause failures when developers use our tools.

You can simply add comments to it for demonstration.

examples/tls_server-rs/ta/Cargo.toml
```diff
@@ -45,3 +49,11 @@ optee-utee-build = { path = "../../../optee-utee-build" }
 panic = "abort"
 lto = false
 opt-level = 3
+
+[package.metadata.optee.ta]
+arch = "aarch64"
+debug = false
+std = true
+uuid-path = "../uuid.txt"
+ta-dev-kit-dir = { aarch64 = "/opt/teaclave/optee/optee_os/out/arm-plat-vexpress/export-ta_arm64", arm = "/opt/teaclave/optee/optee_os/out/arm-plat-vexpress/export-ta_arm32" }
```
ivila: Same as above.

ivila (commented):
LGTM, with 2 minor changes needed.

I think we can merge it after the changes are made. This PR has been open for a long time, so we can ship the tool first, let developers try it, get feedback, and gradually improve it. What's your opinion? @DemesneGH

DemesneGH:
> I think we can merge it after the changes are made. This PR has been open for a long time, so we can ship the tool first, let developers try it, get feedback, and gradually improve it.

@ivila Sure, I think it's good to merge it before the next release (around Jan 30). I will quickly review it these days.

examples/tls_server-rs/ta/Cargo.toml
asir66: I think so. I made this change to fix a build error at the time.
After further investigation, it turned out the error was unrelated, and I forgot to revert the change.
That was my mistake — I will fix it.

cargo-optee/src/main.rs
```diff
@@ -0,0 +1,402 @@
...
+        // CLI provided - resolve relative to current directory
+        std::env::current_dir()?.join(build_config.get_uuid_path())
+    } else {
+        // From metadata or default - resolve relative to project directory
+        project_path.join(build_config.get_uuid_path())
+    };
+
+    // Merge env variables: CLI overrides + metadata env
+    let mut merged_env = build_config.env.clone();
+    merged_env.extend(common.env);
```
DemesneGH: Seems these lines are also resolution, how about moving these into config::BuildConfig::resolve()?

cargo-optee/src/main.rs
```diff
@@ -0,0 +1,402 @@
...
+    let optee_client_export = resolve_path_relative_to_project(
+        &optee_client_export_config,
+        &project_path,
+        PathType::Directory,
+        "OP-TEE client export directory",
+    )?;
+
+    // Merge env variables: CLI overrides + metadata env
+    let mut merged_env = build_config.env.clone();
+    merged_env.extend(common.env);
```
DemesneGH: Same as in execute_ta_command(), consider moving the resolve logic into BuildConfig::resolve()?

cargo-optee/src/main.rs
```diff
@@ -0,0 +1,402 @@
...
+    }
+
+    Ok(())
+}
+
+/// Source cargo environment from a given path
+fn source_cargo_env(env_path: &str) -> bool {
+    if std::path::Path::new(env_path).exists() {
+        std::process::Command::new("bash")
+            .arg("-c")
```
DemesneGH: The `source` run in sub-processes doesn't infect the env of main process.
Regarding `source xxx/.cargo/env` adds the cargo bin into PATH, how about just check if {}/.cargo/bin/cargo exists.
If found, add that directory to std::env directly; If not, prompt the user to install Cargo.

cargo-optee/src/main.rs
```diff
@@ -0,0 +1,402 @@
...
+        component_type,
+        common.arch,
+        Some(common.debug),
+        None, // std not applicable for CA/Plugin
+        None, // ta_dev_kit_dir not needed for CA/Plugin
+        optee_client_export,
+        None, // signing_key not needed for CA/Plugin
+        uuid_path,
+    )?;
+
```
DemesneGH: Now:
user_input => BuildConfig::resolve(ComponentType::Ta) => common build config => handle some Ta-specific params such as uuid
common build config + Ta specific params => TaBuildConfig


How about simplifying like this:
user_input => resolver(ComponentType::Ta) => TaBuildConfig
The resolver handles all params for TA.
Also same for CA: user_input => resolver(ComponentType::Ca) => CaBuildConfig

Is this approach feasible？

cargo-optee/README.md
```diff
@@ -0,0 +1,566 @@
...
+        │                                              │
+        │    - cargo/xargo: Rust compilation           │
+        │    - gcc: Linking with OP-TEE libraries      │
+        │    - objcopy: Symbol stripping               │
+        │    - Python script: TA signing (TA only)     │
+        │                                              │
+        └──────────────────────────────────────────────┘
+```
+
+## Quick Start
```
DemesneGH, line 62: Recommend reorganizing this part:
```
## Quick Start
### Installation
### Quick Build for Hello World // new added, demo of building the example using simple commands
## Configuration System
cargo-optee uses a flexible configuration system...
1. xxxx (link to sub-title)
2. xxxx (link to sub-title)
...
### Project Structure
### Usage Workflows (including future design)
### Build through Cli  // original "Build Commands" part
### Build through metadata
```

docs/emulate-and-dev-in-docker-std.md
```diff
@@ -40,13 +41,19 @@ $ docker run -it --rm \
...
+> Your own projects may organize std files differently.
 
 ## 2. Configuration Management System
 
-The key difference is the **unified configuration system** that allows switching between std/no-std modes and different architectures on demand.
+The key difference is the **unified configuration system** that allows switching between std/no-std 
+modes and different architectures on demand.
 
-### Check Available Configurations
+And cargo-optee is available as an alternative to the original configuration management tool: switch_config.
```
DemesneGH: Also provide the link to cargo-optee README

docs/emulate-and-dev-in-docker-std.md
```diff
@@ -7,7 +7,8 @@ permalink: /trustzone-sdk-docs/emulate-and-dev-in-docker-std.md
 This guide covers the **dev-env with std support** that enables **developing TA using Rust standard library (std)**, 
 compared to the regular no-std environment documented in [emulate-and-dev-in-docker.md](emulate-and-dev-in-docker.md).
 
-The **dev-env with std support** provides a complete setup for building TAs that can use Rust's standard library features like collections, networking, etc.
+The **dev-env with std support** provides a complete setup for building TAs that can use Rust's standard library 
+features like collections, networking, etc.
```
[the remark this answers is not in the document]
DemesneGH: Please rewrap for long lines at 80th characters.
You can use cli tools or plugins in VSCode for automatic formatting.
Same for other modified READMEs.

PR #268, opened 2026-01-09
feat: enable optional static teec linkage

### LINKED ISSUE #267: Support statically linking libteec
optee_client supports compilation of statically linked libraries. But optee-teec-sys *forces* dynamic linkage due to [println!("cargo:rustc-link-lib=dylib=teec")](https://github.com/apache/teaclave-trustzone-sdk/blob/fac9759809308e3ba50f807de322bfd9a98a3c5e/optee-teec/optee-teec-sys/build.rs#L55).

It would be good to have the ability to control whether static or dynamic linkage is used.

I propose following the [conventions used by pkg-config](https://docs.rs/pkg-config/0.3.32/pkg_config/#linking). If the build-time env var `TEEC_STATIC` is present, then static linking will be used, if `TEEC_DYNAMIC` is present, dynamic will be used, and if no such env var is present, it will use the defaults (dynamic).

TheButlah:
Closes #267

### DIFF
--- a/optee-teec/optee-teec-sys/build.rs
+++ b/optee-teec/optee-teec-sys/build.rs
@@ -19,25 +19,29 @@ use std::env::{self, VarError};
 use std::path::PathBuf;
 
 fn main() -> Result<(), VarError> {
-    if !is_feature_enable("no_link")? {
-        link();
+    if !is_feature_enabled("no_link")? {
+        link(is_env_present("TEEC_STATIC")?);
     }
     Ok(())
 }
 
-// Check if feature enabled.
-// Refer to: https://doc.rust-lang.org/cargo/reference/features.html#build-scripts
-fn is_feature_enable(feature: &str) -> Result<bool, VarError> {
-    let feature_env = format!("CARGO_FEATURE_{}", feature.to_uppercase().replace("-", "_"));
-
-    match env::var(feature_env) {
+fn is_env_present(var: &str) -> Result<bool, VarError> {
+    println!("cargo:rerun-if-env-changed={var}");
+    match env::var(var) {
         Err(VarError::NotPresent) => Ok(false),
         Ok(_) => Ok(true),
         Err(err) => Err(err),
     }
 }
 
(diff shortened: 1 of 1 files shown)


DemesneGH:
Thanks for the PR!
Introducing an option for the linking method is a good improvement and the code looks good to me.

However, since this introduces a new implicit environment variable, could you help add some documentation in https://github.com/apache/teaclave-trustzone-sdk/blob/main/docs/advanced-setup.md?
A new section like "## Linkage Method to OP-TEE C Libraries" would be good to explain how to use TEEC_STATIC to control the linking method.

optee-teec/optee-teec-sys/build.rs
```diff
@@ -19,25 +19,29 @@ use std::env::{self, VarError};
 use std::path::PathBuf;
 
 fn main() -> Result<(), VarError> {
-    if !is_feature_enable("no_link")? {
-        link();
+    if !is_feature_enabled("no_link")? {
+        link(is_env_present("TEEC_STATIC")?);
```
ivila, line 23: I think we should check that it’s non-zero rather than just checking for presence, since it could be set as `TEEC_STATIC=0`.

For [example](https://github.com/rust-openssl/rust-openssl/blob/5d012b2ed569aa0e87c299f672d79f2b8eee11be/openssl-sys/build/main.rs#L493-L504):

image

TheButlah: I think it might be unintuitive to diverge at all from pkg-config's semantics. They only check for env var presence, not the 1 vs 0 value. See [here](https://docs.rs/pkg-config/0.3.32/src/pkg_config/lib.rs.html#759).

If you prefer, I could make it panic if set to anything other than 1.

ivila: No changes needed—the current version works fine. 😀

ivila:
@DemesneGH this PR looks ready to go. Are you OK with me merging this now?

PR #278, opened 2026-02-09
fix!: introduce access api and VolatileBuf to safely interact with volatile ParamMemref buffers

### LINKED ISSUE #273: How to use as_memref() safely?
[The docs say](https://teaclave.apache.org/api-docs/trustzone-sdk/optee_utee/struct.Parameter.html#method.as_memref): 
> The caller must ensure that the raw pointer is valid and points to a properly initialized TEE_Param.

But how can we be sure the raw pointer is valid? An adversarial CA could be mutating the buffer concurrently to the TA reading it, and as such maybe this buffer should instead be considered as a volatile pointer. For that reason, constructing either a &[u8] or a &mut [u8] doesn't seem sound unless we can somehow assume the CA is playing nice, which a TA should never assume. I would think that the *only* safe mechanism here is to copy the buffer into the TA's stack or heap, so maybe the API should be reworked to expose a volatile pointer with a safe function to copy it into a slice?

Also, what is to stop the CA from providing a bogus pointer into the TA's own memory, thereby basically getting the TA to read from its own memory space (instead of from CA controlled memory)?

Please advise.

TheButlah:
Fixes https://github.com/apache/teaclave-trustzone-sdk/issues/273

**Summarizing that issue**, the [current api](https://github.com/apache/teaclave-trustzone-sdk/blob/e6511571ef29dc7bd2dd971098027cdcfea86ded/optee-utee/src/parameter.rs#L79-L81) to access a `ParamMemref`'s buffer is unsound when having to account for malicious CAs (and probably benign ones too). The main issue is that since this memory is not synchronized, Linux can change its contents at any time, and therefore it is essentially volatile memory. Constructing a reference to volatile memory is always unsound.

Additionally, OP-TEE *only* allows reads when the memref is an `[in]` or `[inout]` param, and writes when its an `[out]` or `[inout]` param. The original API doesn't check the `ParamType` at all, which means that the end user can easily write to a buffer they can only read from, or read from a buffer they can only write to.

**This PR fixes the issue by introducing two concepts**:
* An "access" api in `access.rs`. This api has traits and zero sized types that allow us to encode the access permissions into the type system via a typestate pattern. This sort of approach is a zero cost abstraction and allows us to check the param_type *only once*, as opposed to every interaction with the memref. It also helps ensure that only the appropriate functions (such as .read() or .write()) are expose for a given access type. The typestate approach is also extremely common in the bare-metal rust ecosystem (see embedded-hal crates), so I figured it isn't overly unfamiliar for this domain of work.
* A `VolatileBuf` in `volatile.rs`, which is a type that ensures that all access to its memory is done through `core::ptr::read_volatile` or `core::ptr::write_volatile`, and leverages the access api to expose only the appropriate read/write functionality.

*A few things reviewers should be aware of:*
* I added a dependency on `bytemuck::Pod`. This was done because I wanted to support *arbitrary* buffers of `T` rather than just `u8`. This was done in an attempt to allow faster accessing of data without needing to repeatedly call deserialization code, as if it were only u8 some sort of deserialization (or subsequent bytemuck cast) would be required. The downside is the additional dependency, but bytemuck is a 1.0 crate and well known and commonly used, and itself has no dependencies. I didn't want to introduce a first-party Pod trait of our own, because it would have less compatibility with the broader ecosystem and would require us to then implement bytemuck-style derive macros for user-defined types.
* I *didn't* use the [volatile crate](https://docs.rs/volatile) because so much of the crate's surface area requires nightly compiler features. Additionally, I found the API limiting in some ways, and too complicated to easily use in others. It should be noted that the volatile crate *also* utilizes a type-state access api.
* This is a breaking change. It breaks the example code too, and I wanted to hold off on fixing that to keep the diff small and let us discuss first, and align on the approach. Once I get a green light that this direction is desired, I'll fix the rest.

I am definitely open to workshopping this submission, please let me know your thoughts and I will adjust the PR accordingly. Thanks for your time :+1:

### DIFF
--- a/examples/serde-rs/ta/src/main.rs
+++ b/examples/serde-rs/ta/src/main.rs
@@ -75,7 +75,7 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
 
             // update size of output buffer
             if p.set_updated_size(len).is_err() {
-                unreachable!()
+                unreachable!("already made sure the buffer has enough capacity")
             }
 
             // Prints serialized = {"x":1,"y":2}
--- a/examples/serde-rs/ta/src/main.rs
+++ b/examples/serde-rs/ta/src/main.rs
@@ -70,7 +70,12 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
             let len = bytes.len();
             if len > buffer.capacity() {
                 trace_println!("Buffer too small, cannot copy all bytes");
-                p.request_more_capacity(len).expect("infallible");
+                // by convention, TAs can hint to a CA that more capacity is
+                // needed. Note that the CA in this example doesn't make use of
+                // this hint.
+                if let Err(_) = p.request_more_capacity(len) {
+                    unreachable!()
+                }
                 return Err(ErrorKind::BadParameters.into());
             }
 
@@ -78,7 +83,9 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
             buffer.copy_from(bytes);
 
             // update size of output buffer
-            p.set_updated_size(len).expect("infallible");
+            if let Err(_) = p.set_updated_size(len) {
+                unreachable!()
+            }
 
             // Prints serialized = {"x":1,"y":2}
             trace_println!("serialized = {}", serialized);
--- a/examples/serde-rs/ta/src/main.rs
+++ b/examples/serde-rs/ta/src/main.rs
@@ -80,7 +80,7 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
             }
 
             // Copy the serialized JSON string into the buffer.
-            buffer.copy_from(bytes);
+            buffer.copy_from(bytes)?;
 
             // update size of output buffer
             if let Err(_) = p.set_updated_size(len) {
--- a/examples/serde-rs/ta/src/main.rs
+++ b/examples/serde-rs/ta/src/main.rs
@@ -68,16 +68,7 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
 
             // Ensure the buffer is large enough to hold the serialized data.
             let len = bytes.len();
-            if len > buffer.capacity() {
-                trace_println!("Buffer too small, cannot copy all bytes");
-                // by convention, TAs can hint to a CA that more capacity is
-                // needed. Note that the CA in this example doesn't make use of
-                // this hint.
-                if p.request_more_capacity(len).is_err() {
-                    unreachable!()
-                }
-                return Err(ErrorKind::BadParameters.into());
-            }
+            p.ensure_capacity(len)?;
 
             // Copy the serialized JSON string into the buffer.
             buffer.copy_from(bytes)?;
--- a/examples/serde-rs/ta/src/main.rs
+++ b/examples/serde-rs/ta/src/main.rs
@@ -24,7 +24,7 @@ use alloc::string::String;
 use optee_utee::{
     ta_close_session, ta_create, ta_destroy, ta_invoke_command, ta_open_session, trace_println,
 };
-use optee_utee::{Error as TeeError, ErrorKind, Parameters, Result};
+use optee_utee::{ErrorKind, Parameters, Result};
 use proto::{Command, Point};
 
 #[ta_create]
@@ -55,7 +55,7 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
     match Command::from(cmd_id) {
         Command::DefaultOp => {
             let mut p = unsafe { params.0.as_memref()? }.output()?;
-            let mut buffer = p.buffer().ok_or(TeeError::new(ErrorKind::ShortBuffer))?;
+            let mut buffer = p.buffer()?;
             let point = Point { x: 1, y: 2 };
 
             // Convert the Point to a JSON string.
--- a/examples/serde-rs/ta/src/main.rs
+++ b/examples/serde-rs/ta/src/main.rs
@@ -24,7 +24,7 @@ use alloc::string::String;
 use optee_utee::{
     ta_close_session, ta_create, ta_destroy, ta_invoke_command, ta_open_session, trace_println,
 };
-use optee_utee::{ErrorKind, Parameters, Result};
+use optee_utee::{Error as TeeError, ErrorKind, Parameters, Result};
 use proto::{Command, Point};
 
 #[ta_create]
@@ -54,8 +54,8 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
     trace_println!("[+] TA invoke command");
     match Command::from(cmd_id) {
         Command::DefaultOp => {
-            let mut p = unsafe { params.0.as_memref()? };
-            let buffer = p.buffer();
+            let mut p = unsafe { params.0.as_memref()? }.output()?;
+            let mut buffer = p.buffer().ok_or(TeeError::new(ErrorKind::ShortBuffer))?;
             let point = Point { x: 1, y: 2 };
 
             // Convert the Point to a JSON string.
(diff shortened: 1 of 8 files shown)


optee-utee/src/parameter.rs
```diff
@@ -171,6 +291,20 @@ pub enum ParamType {
     MemrefInout = 7,
 }
 
+impl Display for ParamType {
```
DemesneGH: Recommend using strum_macros::Display

optee-utee/src/parameter.rs
```diff
@@ -69,27 +77,137 @@ impl<'parameter> ParamValue<'parameter> {
...
+            raw: self.raw,
+            param_type: self.param_type,
+            capacity: self.capacity,
+            _phantom_param: marker::PhantomData,
+            _phantom_access: marker::PhantomData,
+        })
+    }
+
+    /// Checks that this param is a [`ParamType::MemrefInput`] and casts access to [`Read`].
+    pub fn in_(self) -> Result<ParamMemref<'parameter, Read>, InvalidAccessErr<Self, Read>> {
```
DemesneGH: I see `in` is a reserved keyword in Rust, but `in_()` is a little awkward for user interface. How about naming as `input()`, `output()`?

TheButlah:
OK I've addressed the feedback so far, let me know what you think :+1: 
I updated the serde example

examples/serde-rs/ta/src/main.rs
```diff
@@ -54,8 +54,8 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
     trace_println!("[+] TA invoke command");
     match Command::from(cmd_id) {
         Command::DefaultOp => {
-            let mut p = unsafe { params.0.as_memref()? };
-            let buffer = p.buffer();
+            let mut p = unsafe { params.0.as_memref()? }.output()?;
+            let mut buffer = p.buffer().ok_or(TeeError::new(ErrorKind::ShortBuffer))?;
```
DemesneGH: Seems we always need to cast the option to error, how about buffer() directly returns Result?

examples/serde-rs/ta/src/main.rs
```diff
@@ -68,16 +68,17 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
 
             // Ensure the buffer is large enough to hold the serialized data.
             let len = bytes.len();
-            if len > buffer.len() {
+            if len > buffer.capacity() {
                 trace_println!("Buffer too small, cannot copy all bytes");
+                p.request_more_capacity(len).expect("infallible");
```
DemesneGH: We should avoid panic in TAs, could you revise the interface for better error handling?
Same in line81.

examples/serde-rs/ta/src/main.rs
```diff
@@ -68,16 +68,17 @@ fn invoke_command(cmd_id: u32, params: &mut Parameters) -> Result<()> {
...
                 return Err(ErrorKind::BadParameters.into());
             }
 
             // Copy the serialized JSON string into the buffer.
-            buffer[..len].copy_from_slice(bytes);
+            buffer.copy_from(bytes);
```
DemesneGH: Suggest to handle the result of `copy_from()`

optee-utee/src/parameter.rs
```diff
@@ -127,9 +246,11 @@ impl Parameter {
         match self.param_type {
             ParamType::MemrefInout | ParamType::MemrefInput | ParamType::MemrefOutput => {
                 Ok(ParamMemref {
-                    raw: &mut (*self.raw).memref,
+                    raw: NonNull::new_unchecked(addr_of_mut!((*self.raw).memref)),
                     param_type: self.param_type,
-                    _marker: marker::PhantomData,
+                    capacity: unsafe { *self.raw }.memref.size,
```
DemesneGH: An internal `unsafe` block isn't strictly necessary because the entire function is already marked `unsafe`

examples/serde-rs/ta/src/main.rs
TheButlah: ok, let me know if `unreachable!()` is more clear. The point is that this code should never panic.

optee-utee/src/parameter.rs
TheButlah: Done. BTW 2024 edition turns on warnings by default if the `unsafe` keyword is not used explicitly. It might be good for teaclave to `#![forbid(unsafe_op_in_unsafe_fn)]` by default.

examples/serde-rs/ta/src/main.rs
DemesneGH: I'm a bit confused about the logic here:

`request_more_capacity()` could return error according its inner logic, right?
Regarding that, it should raise the error to the caller.

Current code:
```
                if p.request_more_capacity(len).is_err() {
                    unreachable!()
                }
```
means it should never return error?

optee-utee/src/parameter.rs
```diff
@@ -185,3 +306,101 @@ impl From<u32> for ParamType {
...
+    }
+}
+
+impl CoreError for BiggerThanCapacityErr {}
+
+impl From<BiggerThanCapacityErr> for crate::Error {
+    fn from(_value: BiggerThanCapacityErr) -> Self {
+        crate::Error::new(ErrorKind::Overflow)
+    }
+}
```
DemesneGH, line 332: Seems these new error types are only used for printing error messages before being converted into a standard optee Error?

To keep the SDK simple, why not use `trace_println!` at the call site and eliminate these extra types? I believe the code would be much clean and simple.

BTW, you can use `ErrorKind::Overflow.into()` to convert directly to the OP-TEE Error.

examples/serde-rs/ta/src/main.rs
TheButlah: the request_more_capacity() function will error if the requested capacity `len` is less than or equal to the current capacity. We already checked this on line 71, which is why the function will never return an error. To indicate the intent more clearly, I used `unreachable!()` although normally when I write code, I just `.expect("infallible")`. We could bubble the error up with `?` but imo obscures the fact that this error will never happen in the first place.

let me know how to proceed.

optee-utee/src/parameter.rs
TheButlah: > Seems these new error types are only used for printing error messages before being converted into a standard optee Error?

I provide the `From` conversion to make the example code more terse, but I believe there is value in more clearly communicating what the nature of the error is (along with a richer error message). In production code for example, we actually wrap all of our errors with `anyhow` and then right before we return to optee, we log that more comprehensive error out, and having the ability to clearly see in logs the chain of errors and their descriptions (without having to clutter every call site with tons of if lets and printlns), is very valuable.

There are other reasons too. For example, in the system we use at work, our serial is not accessible therefore if we want to see what errors happened, we need to return the error to the CA. So we can't rely on `trace_*!()` macros. If we used only `ErrorKind` we would have *very* little information necessary to debug the system, since so little information can be encoded. Instead, it makes sense to have richer error types that can be returned to the CA. Additionally, trace_println!() is extremely slow and becomes a performance bottleneck very fast.

The way this PR does it is the best of both worlds: it provides the conversion for those that just want a `?` and tersely writing code, but it retains the structured error for those that need/want it.

Side note: some of these error types are more than just richer error messages. The error returned when converting access for example, contains the original type (since the function takes `self`) so that if the conversion fails, the caller doesn't drop data and can try again with a different access type.

DemesneGH: Sounds good, thanks

examples/serde-rs/ta/src/main.rs
DemesneGH: Suggestions on `NotBiggerThanCapacityErr`:
- the name is a bit confusing, how about `ShortBufferErr` or any other you prefer
- It should be converted into `ErrorKind::ShortBuffer`, as TA notifies CA that it needs more capacity

Suggestions on `request_more_capacity()`, how about changing to:
```
pub fn ensure_capacity(&mut self, required_len: usize) -> Result<(), ShortBufferErr> {
    if required_len > self.capacity {
        let memref = unsafe { self.raw.as_mut() };
        memref.size = required_len;
        return Err(ShortBufferErr);
    }
    Ok(())
}
```
In the example:
```
let len = bytes.len();
// no need to manually check the len
p.ensure_capacity(len)?
```

TheButlah: I think that approach is a big improvement! I've updated the PR

TheButlah:
My concern would be, what if someone *doesn't* want to update the capacity?

ivila:
> I believe the Rust SDK should provide primitives that are as safe as possible. In our context, checking capacity, copying data, and updating the size should be treated as a single atomic operation to maintain consistency. Decoupling them only creates more opportunities for logic bugs (like forgetting to update the size) and leads to redundant error handling.
> 
> > ~what if someone _doesn't_ want to update the capacity?~
> 
> I cannot think of a valid use case for skipping the update... If the capacity is insufficient, why not updating the capacity to inform the client？
> 
> > maybe we _should_ just always ensure capacity and set updated size when doing copy_from?
> 
> It also makes sense to me. To keep the API explicit, we may need a more descriptive name than just copy_from to indicate it handles the full lifecycle (check, copy, and size update).
> 
> Also requesting feedback from @ivila for this API design (once he is available). Since we are heading into the Chinese New Year holiday, our responses would be slower than usual. Thanks for your patience!

Sorry for the late response — I must have missed the GitHub notification. I don’t have any other comments; only a few parts of the code need to be updated (for example, we should use core::error directly rather than renaming it to CoreError, just check previous comments).😄

ivila:
Already Done by #317

PR #281, opened 2026-02-16
Fix: pin uuid version for rust 1.80.0 compatibility

aimir:
The new `uuid@1.21.0` supports rust 1.85 and above, so this pins it to `1.20.0`


DemesneGH:
@aimir Could you also pin the version here: https://github.com/apache/teaclave-trustzone-sdk/blob/main/projects/web3/eth_wallet/host/Cargo.toml

PR #285, opened 2026-03-10
build: update MSRV to nightly-2025-12-11

DemesneGH:
Changes:
- Updated `setup_std_dependencies.sh` to clone the pinned Rust source and apply patches from teaclave-crates https://github.com/apache/teaclave-crates/pull/10. After that PR merged the clone source should be updated in `setup_std_dependencies.sh`.
- Resolved compilation and Clippy errors introduced by the newer compiler version.
- Continued use of Xargo for this transition, with `XARGO_RUST_SRC` pointing to the new `library/` workspace.

Status:
- Basic CI passes.
- Temporarily excluded `std-for-dev-docker` and `cargo-optee-*` for this PR. Since the build toolchain will be overhauled in the next step, these environments will be updated accordingly at that time.

Next Steps:
- Migrate from Xargo to `cargo -Z build-std` as the primary build tool, as Xargo is outdated and no longer updated.
- Fix remaining CI errors in the excluded components.

### DIFF
--- a/examples/message_passing_interface-rs/ta/Cargo.lock
+++ b/examples/message_passing_interface-rs/ta/Cargo.lock
@@ -61,12 +61,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7581282928bc99698341d1de7590964c28db747c164eaac9409432a3eaed098a"
 
 [[package]]
-name = "litemap"
-version = "0.7.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4ee93343901ab17bd981295f2cf0026d4ad018c7c31ba84549a4ddbb47a45104"
-
-[[package]]
 name = "log"
 version = "0.4.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -122,23 +116,19 @@ dependencies = [
 name = "optee-utee-build"
 version = "0.8.0"
 dependencies = [
- "litemap",
  "prettyplease",
  "proc-macro2",
  "quote",
  "syn",
- "uuid 1.20.0",
- "zerofrom",
+ "uuid 1.22.0",
 ]
 
 [[package]]
 name = "optee-utee-macros"
 version = "0.8.0"
 dependencies = [
- "litemap",
  "quote",
  "syn",
- "zerofrom",
 ]
 
 [[package]]
@@ -292,9 +282,9 @@ checksum = "bc5cf98d8186244414c848017f0e2676b3fcb46807f6668a97dfe67359a3c4b7"
 
 [[package]]
 name = "uuid"
-version = "1.20.0"
+version = "1.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ee48d38b119b0cd71fe4141b30f5ba9c7c5d9f4e7a3a8b4a674e4b6ef789976f"
+checksum = "a68d3c8f01c0cfa54a75291d83601161799e4a89a39e0929f4b0354d88757a37"
 dependencies = [
  "js-sys",
  "wasm-bindgen",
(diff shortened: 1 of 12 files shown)


examples/client_pool-rs/ta/Makefile
ivila: We should check BUILDER after it's been declared.

Maybe move the changes after line 39.

examples/serde-rs/ta/Cargo.toml
```diff
@@ -15,6 +15,8 @@
 # specific language governing permissions and limitations
 # under the License.
 
+cargo-features = ["panic-immediate-abort"]
+
```
ivila: Perhaps we should wrap the hotfix functions (such as `rust_eh_personality` and `_Unwind_Resume`) behind a `no_std_abort_fix` feature gate in `optee-utee` crate. This avoids relying on unstable Cargo features and build-std, which force us to use the nightly toolchain even at `no_std`.

setup_std_dependencies.sh
```diff
@@ -25,12 +25,19 @@ cd "$(dirname "$0")"
...
+	rustup toolchain install "$XARGO_BOOTSTRAP_TOOLCHAIN" --profile minimal
+	cargo +"$XARGO_BOOTSTRAP_TOOLCHAIN" install xargo --locked
+fi
 
 ##########################################
-# initialize submodules: rust / libc
-RUST_BRANCH=optee-xargo
-LIBC_BRANCH=optee
+# initialize submodules: rust / libc / patches
+RUST_COMMIT=01f6ddf7588f42ae2d7eb0a2f21d44e8e96674cf  # rust 1.93.1 stable release
```
ivila: I think using a more readable branch name would be better, such as `rust_1.93.1_patch`

setup_std_dependencies.sh
```diff
@@ -39,11 +46,22 @@ fi
 
 mkdir rust && cd rust
 
-git clone --depth=1 -b $RUST_BRANCH https://github.com/DemesneGH/rust.git && \
+# Clone official Rust source at specific commit
+git clone https://github.com/rust-lang/rust.git && \
```
ivila: Omitting --depth=1 will waste a lot of bandwidth.

setup_std_dependencies.sh
DemesneGH: For the previous approach we maintain the fork of whole repo, so we can define a branch name; The current approach pins the specific commit in upstream Rust, we only maintain the patch in `teaclave-crates`. Seems we cannot have a branch name like that. Any suggestions?

---
Update: change to use the tag name with `depth=1`

aarch64-unknown-optee.json
```diff
@@ -1,6 +1,6 @@
 {
     "arch": "aarch64",
-    "data-layout": "e-m:e-i8:8:32-i16:16:32-i64:64-i128:128-n32:64-S128",
+    "data-layout": "e-m:e-p270:32:32-p271:32:32-p272:64:64-i8:8:32-i16:16:32-i64:64-i128:128-n32:64-S128-Fn32",
```
ivila, line 3: There are duplicate files in the cargo-optee directory. Should these be updated as well, or would it be better to remove them entirely?

examples/serde-rs/ta/Makefile
```diff
@@ -26,11 +26,6 @@ LINKER_CFG := target.$(TARGET).linker=\"$(CROSS_COMPILE)gcc\"
 # fix for the error: "unwinding panics are not supported without std" reported by clippy
 # Set panic=abort for std and no-std
 RUSTFLAGS := -C panic=abort
-ifeq ($(BUILDER),xargo)
-EXTRA_FLAGS =
-else
-EXTRA_FLAGS = -Z build-std=core,alloc -Z build-std-features=panic_immediate_abort
```
ivila, line 32: `cargo-optee` adds the `build-std` flag by default; I think we should remove that as well.

https://github.com/apache/teaclave-trustzone-sdk/blob/925a5d4b268c4e2e23255d108d111be7feb48752/cargo-optee/src/ta_builder.rs#L362-L368

aarch64-unknown-optee.json
DemesneGH: The modification of `cargo-optee` is planned in the follow-up PRs, after this PR merged I will remove the Xargo and update our building system. To reduce the effort of multiple modifications I plan to adjust the cargo-optee after that.
For this PR, I suggest we omit the CI errors of `cargo-optee` and `std-dev-docker`.

PR #296, opened 2026-04-29
optee-teec: refactor plugin support

ivila:
* Introduce `optee-teec-plugin-bindgen` crate to generate plugin bindings for plugin developers, replacing `optee-teec-macros`.
* Refactor `optee-teec::PluginParameters` and add additional methods.
* Update `optee-teec-sys::PluginMethod` to latest version.

With these changes, plugin development now only requires adding `optee-teec-plugin-bindgen` as a build dependency and including the generated artifact in `lib.rs`.

## Why not use a proc macro?

I have thought about implementing a `declare_supp_plugin` macro like this[demo](https://github.com/apache/teaclave-trustzone-sdk/compare/main...ivila:teaclave-trustzone-sdk:opt_plugin_0427):
```Rust
declare_supp_plugin!(
    name: "my_plugin",
    uuid: "ef620757-fa2b-4f19-a1c4-6e51cfe4c0f9",
    init: my_init,
    invoke: my_invoke,
);
```
Developers can use it like:
```rust
use optee_teec::{declare_supp_plugin, PluginParameters};

fn plugin_init() -> optee_teec::Result<()> {
    println!("*plugin*: init, version: {}", env!("CARGO_PKG_VERSION"));
    Ok(())
}

fn plugin_invoke(_: &mut PluginParameters) -> optee_teec::Result<()> {
    println!("*plugin*: invoke");
    Ok(())
}

declare_supp_plugin!(
    name: "syslog",
    uuid: "ef620757-fa2b-4f19-a1c4-6e51cfe4c0f9",
    init: plugin_init,
    invoke: plugin_invoke,
);
```
The main issue is that developers may provide the UUID through a variable. Due to the limitations of proc macros, we cannot resolve the value of that variable during macro expansion. As a result, the UUID would need to be parsed and converted at runtime.
```Rust
declare_supp_plugin!(
    name: "my_plugin",
    uuid: proto::PLUGIN_UUID,  // proc_macro cannot resolve the value at compile time.
    init: my_init,
    invoke: my_invoke,
);
```

However, this introduces another problem: what happens if UUID parsing fails? Triggering a panic in a shared library is not a good idea and should be avoided.

### DIFF
--- a/crates/Cargo.toml
+++ b/crates/Cargo.toml
@@ -19,6 +19,7 @@
 resolver = "3"
 members = [
     "optee-teec",
+    "optee-teec-build",
     "optee-teec-macros",
     "optee-teec-sys",
     "optee-teec-systest",
@@ -40,6 +41,8 @@ description = "TEE internal core API."
 edition = "2024"
 
 [workspace.dependencies]
+optee-teec = { version = "0.8.0", path = "optee-teec" }
+optee-teec-build = { version = "0.8.0", path = "optee-teec-build" }
 optee-teec-macros = { version = "0.8.0", path = "optee-teec-macros" }
 optee-teec-sys = { version = "0.8.0", path = "optee-teec-sys" }
 optee-utee = { version = "0.8.0", path = "optee-utee" }
@@ -52,8 +55,10 @@ uuid = { version = "1.23", default-features = false }
 hex = { version = "0.4", default-features = false, features = ["alloc"] }
 quote = "1.0"
 syn = { version = "2.0", features = ["full"] }
+proc-macro2 = "1.0.92"
 libc = "0.2" 
 rand = "0.10"
 once_cell = "1.20.2"
 serde = { version = "1.0.228" }
 serde_json = { version = "1.0.149" }
+log = "0.4.29"
--- a/crates/optee-teec-build/Cargo.toml
+++ b/crates/optee-teec-build/Cargo.toml
@@ -16,8 +16,8 @@
 # under the License.
 
 [package]
-name = "optee-utee-build"
-description = "Build tool for TA"
+name = "optee-teec-build"
+description = "build tool for teec"
 version.workspace = true
 authors.workspace = true
 license.workspace = true
@@ -27,6 +27,5 @@ edition.workspace = true
 [dependencies]
 quote.workspace = true
 proc-macro2.workspace = true
-syn.workspace = true
-prettyplease = "0.2.25"
 uuid.workspace = true
+syn.workspace = true
--- a/crates/optee-teec-build/src/lib.rs
+++ b/crates/optee-teec-build/src/lib.rs
@@ -15,13 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-#[no_mangle]
-pub static mut plugin_method: optee_teec::PluginMethod = optee_teec::PluginMethod {
-    name: plugin_name.as_ptr(),
-    uuid: PLUGIN_UUID_STRUCT,
-    init: _plugin_init,
-    invoke: _plugin_invoke,
-};
+mod plugin;
+pub use uuid;
 
-#[no_mangle]
-pub static plugin_name: &[u8] = b"syslog\0";
+pub use plugin::{DEFAULT_INIT_FN_NAME, DEFAULT_INVOKE_FN_NAME, PluginConfig};
--- a/crates/optee-teec-build/src/plugin.rs
+++ b/crates/optee-teec-build/src/plugin.rs
@@ -1,0 +1,120 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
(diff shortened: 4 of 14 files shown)


examples/supp_plugin-rs/plugin/build.rs
```diff
@@ -15,32 +15,14 @@
...
-    clockSeqAndNode: {:#x?},
-}};",
-        time_low, time_mid, time_hi_and_version, clock_seq_and_node
-    )?;
+fn main() -> anyhow::Result<()> {
+    let config = optee_teec_plugin_bindgen::Config::new(
+        "syslog",
+        uuid::Uuid::parse_str(proto::PLUGIN_UUID)?,
+        "init",
+        "invoke",
```
DemesneGH: I see why we're avoiding macros, but using strings for function names is a bit weird. 
Since most users don't need custom names anyway, how about  using defaults like `plugin_init` and `plugin_invoke`? Then the default constructor only requires the name and UUID. This would clean up the API and eliminate potential typos in `build.rs`.

We can also provide one more constructor that accepts the customized fn name if you'd like to.

ivila: Sounds good!😃

ivila: I have restored `optee-teec-macros` and am considering using macros to generate the injected functions automatically. What do you think about this approach?

With this solution, developers would apply macros to their init and invoke functions:
```rust
// lib.rs
#[derive_raw_plugin_init]
fn init() -> optee_teec::Result<()> {
    Ok(())
}

#[derive_raw_plugin_invoke]
fn invoke(_: &mut optee_teec::PluginParameters) -> optee_teec::Result<()> {
    Ok(())
}

// build.rs
fn main() -> anyhow::Result<()> {
    Config::new(Uuid::parse_str(proto::PLUGIN_UUID)?)
        .with_name("syslog")
        .build()?;

    Ok(())
}
```
If they need to customize the generated raw function name, it could work like this:
```rust
// lib.rs
#[derive_raw_plugin_init(raw_name = "raw_func_name")]
fn init() -> optee_teec::Result<()> {
    Ok(())
}

#[derive_raw_plugin_invoke]
fn invoke(_: &mut optee_teec::PluginParameters) -> optee_teec::Result<()> {
    Ok(())
}

// build.rs
fn main() -> anyhow::Result<()> {
    Config::new(Uuid::parse_str(proto::PLUGIN_UUID)?)
        .with_name("syslog")
        .with_init_fn_name("raw_func_name")
        .build()?;

    Ok(())
}
```

ivila: Here's another idea I am considering: introducing a dedicated UUID conversion crate (for example, `optee-teec-build`) that converts a UUID string into `raw::TEEC_UUID` during build.rs.

With this approach, `build.rs` would generate the UUID constant, and `lib.rs` would simply include the generated file and use the `declare_supp_plugin` macro to declare the plugin.

```rust
// build.rs
const OUTPUT_VAR_NAME: &str = "TA_PLUGIN_UUID";
const OUTPUT_FILENAME: &str = "ta_plugin_uuid.rs";

optee_teec_build::convert_uuid(
    uuid_str,
    OUTPUT_VAR_NAME,
    OUTPUT_FILENAME,
)?;

Ok(())
```

```rust
// lib.rs
include!(concat!(env!("OUT_DIR"), "/ta_plugin_uuid.rs"));

declare_supp_plugin!(
    name: "syslog",
    uuid: TA_PLUGIN_UUID,
    init: plugin_init,
    invoke: plugin_invoke,
);
```

This approach keeps the UUID parsing and conversion logic in the build phase, while keeping the plugin declaration in lib.rs clean and declarative.

DemesneGH: Thanks for your insightful proposals! The both approaches make sense to me technically. 

However, considering consistency with the existing TA development flow, how about:

1. Use attribute macros (like `#[plugin_init]` and `#[plugin_invoke]`) from approach 1: This mirrors the `#[ta_create]` pattern that developers are already familiar with. Also keep the macro name in similar pattern if possible.

2. Introduce `optee-teec-build `and `PluginConfig`: This provides a symmetric API to `TaConfig`. It hides the complexity of UUID parsing and code generation (e.g., `optee_teec_build::build_plugin(plugin_config)`).

This alignment makes our API design more intuitive, if users know how to write a TA, they already know how to write a Plugin.

Edit: since we already have Config related implementation in `optee-teec-plugin-bindgen`, how about directly naming this to a sub module of `optee-teec-build`?

ivila: > Thanks for your insightful proposals! The both approaches make sense to me technically.
> 
> However, considering consistency with the existing TA development flow, how about:
> 
> 1. Use attribute macros (like `#[plugin_init]` and `#[plugin_invoke]`) from approach 1: This mirrors the `#[ta_create]` pattern that developers are already familiar with. Also keep the macro name in similar pattern if possible.
> 2. Introduce `optee-teec-build `and `PluginConfig`: This provides a symmetric API to `TaConfig`. It hides the complexity of UUID parsing and code generation (e.g., `optee_teec_build::build_plugin(plugin_config)`).
> 
> This alignment makes our API design more intuitive, if users know how to write a TA, they already know how to write a Plugin.
> 
> Edit: since we already have Config related implementation in `optee-teec-plugin-bindgen`, how about directly naming this to a sub module of `optee-teec-build`?

Changes made.

DemesneGH:
LGTM, thanks!

PR #304, opened 2026-06-03
utee: export define_property_key and update set_key_2

ldshyu:
* export define_property_key to support getting TA defined properties
* set_key_2 is also needed on derive key, move the implementation to OperationHandle

### DIFF
--- a/crates/optee-utee/src/crypto_op.rs
+++ b/crates/optee-utee/src/crypto_op.rs
@@ -223,6 +223,19 @@ impl OperationHandle {
         }
     }
 
+    fn set_key_2<T: GenericObject, D: GenericObject>(
+        &self,
+        object1: &T,
+        object2: &D,
+    ) -> Result<()> {
+        match unsafe {
+            raw::TEE_SetOperationKey2(self.handle(), *object1.as_raw_ref(), *object2.as_raw_ref())
+        } {
+            raw::TEE_SUCCESS => Ok(()),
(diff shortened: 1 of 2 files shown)


ivila:
for
> export define_property_key to support getting TA defined properties

Do you mean custom TA properties? From what I can see, there are already many TA-defined properties implemented in the codebase. For example: https://github.com/apache/teaclave-trustzone-sdk/blob/b3154cb8a4490c38fd2b547bf5fdbc960daf3666/crates/optee-utee/src/property.rs#L278-L470

Could you clarify which additional properties this change is intended to support?

ldshyu:
> for
> 
> > export define_property_key to support getting TA defined properties
> 
> Do you mean custom TA properties? From what I can see, there are already many TA-defined properties implemented in the codebase. For example:
> 
> https://github.com/apache/teaclave-trustzone-sdk/blob/b3154cb8a4490c38fd2b547bf5fdbc960daf3666/crates/optee-utee/src/property.rs#L278-L470
> 
> Could you clarify which additional properties this change is intended to support?

OPTEE allows user TA to define their own properties.
For example, see `TA_CURRENT_TA_EXT_PROPERTIES` in https://github.com/OP-TEE/optee_test/blob/master/ta/os_test/include/user_ta_header_defines.h
By exporting the macro, 
I can do
```
MyPropTrue.get()?
...
define_property_key!(MyPropTrue, CurrentTa, "myprop.true", bool);
```
in user TA.

PR #305, opened 2026-06-04
docs: add security model describing trust boundaries and TCB

m4sterchain:
## What

Adds `docs/security-model.md`, a document describing the SDK's security model. The repository previously had no SDK-level security/trust-model documentation — only `SECURITY.md` (a vulnerability-reporting policy) and a per-demo "Security Assumptions" section in `eth_wallet`.

## Why

TrustZone development differs from ordinary application development: code is partitioned into an untrusted Normal World and a trusted Secure World, and the boundary between them is the central security property. A single document that maps this trust model onto the codebase helps both human developers and automated/LLM security reviewers reason correctly — in particular, to focus on the real trust boundary and avoid false positives on code where the concern does not apply.

## Contents

`docs/security-model.md` covers:

1. **Trust model** — TCB, adversary capabilities (full Normal-World root: arbitrary parameters, attacker-chosen buffer pointers/lengths, concurrent mutation of shared memory), and out-of-scope threats.
2. **The trust boundary** — the TA entry point; per-parameter-type trust analysis (`Value*` vs. shared-memory `Memref*`); and the invariants TA boundary code must enforce (validate `param_types`, treat every memref byte as adversarial, bound lengths, copy-then-validate to avoid TOCTOU/double-fetch, don't leak via output buffers, fail closed).
3. **Trust-posture map of the repository** — a table mapping each path to its world/role and what to scrutinize.
4. **Storage, secrets, and other assumptions** — secure-storage confidentiality vs. rollback/availability, RPMB, secure UI.
5. **Dependencies and the supply chain** — a TA's full transitive crate tree runs inside the TEE and is part of the TCB; `no-std` vs `std` as a TCB-size decision; software crypto and entropy considerations; build-time supply-chain risk; and vetting recommendations.
6. **Guidance for automated / LLM security review** — where real findings concentrate, expected non-findings (false positives), and an explicit instruction to also review the TA's dependency tree when tooling allows.

Also links the new doc from `SECURITY.md` and the docs index (`docs/README.md`).

This is documentation only; no code or behavior changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,5 +1,11 @@
 # Security Policy
 
+## Security Model
+
+For the SDK's trust model, trust boundaries (Normal World vs. Secure World), and
+security assumptions — including guidance for developers and automated security
+reviewers — see [docs/security-model.md](docs/security-model.md).
+
 ## Reporting a Vulnerability
 
 We take a very active stance in eliminating security problems in Teaclave. We
--- a/docs/README.md
+++ b/docs/README.md
@@ -8,6 +8,9 @@ permalink: /trustzone-sdk-docs
 * [Quick Emulation And Development in Docker](emulate-and-dev-in-docker.md)
 * [Developing TAs with Rust Standard Library in Docker](emulate-and-dev-in-docker-std.md)
 
+## Security
+* [Security Model](security-model.md)
+
 ## Development
 * [TA Development Modes](ta-development-modes.md)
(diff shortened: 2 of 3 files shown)


m4sterchain:
@DemesneGH @ivila I am planning to merge these docs changes tomorrow, please take a look if any correction needed when you are available.

PR #308, opened 2026-06-24
optee-utee: refactor parameter

ivila:
Introduce a typed parameter API for optee-utee, migrate examples/projects to it, fix parameter type decoding on the client side, and add migration documentation for TA developers.

  - Replace hand-written ParamType conversion with num_enum derives.
  - Add typed TA parameter support in optee-utee:
      - ParameterValue{Input,Output,Inout}
      - ParameterMemref{Input,Output,Inout}
      - ParameterNone
      - ParameterAny
      - FromRawParameter<'_> and FromRawParameters<'_>
      - ParametersAny<'_> and ParametersNone

  - Move the previous raw-pointer parameter API into optee_utee::deprecated.
  - Update TA entry-point macros to convert raw OP-TEE parameters through FromRawParameters.
  - Add optee_utee::prelude re-exports for TA macros and typed parameter helpers.
  - Migrate examples and projects/web3/eth_wallet TA code to the typed parameter API.
  - Fix optee-teec ParamTypes::into_flags decoding so returned operation parameters are decoded from the correct nibble.
  - Add a TA-focused migration guide for moving from Parameters to typed parameters.
  - Link the migration guide from docs/README.md.


examples/aes-rs/ta/src/main.rs
```diff
@@ -149,31 +151,28 @@ pub fn set_aes_key(aes: &mut AesCipher, params: &mut Parameters) -> Result<()> {
...
         return Err(ErrorKind::BadParameters.into());
     }
 
     trace_println!("[+] TA tries to update ciphers!");
 
-    let tmp_size = aes.cipher.update(input, output)?;
-    param1.set_updated_size(tmp_size);
+    let tmp_size = aes
+        .cipher
+        .update(input.as_slice(), unsafe { output.as_mut_slice() })?;
```
DemesneGH: Since we have eliminated most `unsafe` in the example code, can we also avoid using  `unsafe` here?
The interfaces seems a bit not user-friendly when writing back to output.

crates/optee-utee/src/parameter/memref.rs
```diff
@@ -0,0 +1,208 @@
...
+///
+/// Implemented by [`ParameterMemrefInput`] and [`ParameterMemrefInout`].
+pub trait ParameterMemrefRead {
+    /// Returns the buffer contents as a byte slice.
+    ///
+    /// For `ParameterMemrefInput` the length is the original buffer size as
+    /// supplied by the host. For `ParameterMemrefInout` the length is the
+    /// full buffer capacity, not the number of valid bytes (which may have
+    /// been updated by a prior write).
+    fn as_slice(&self) -> &[u8];
```
DemesneGH: `as_slice()` might be a little confusing here. In Rust, readers often assume `as_slice()` returns the valid range rather than the full capacity.
Can we clarify it by using a different name, or other ways you prefer?

examples/aes-rs/ta/src/main.rs
ivila: making `as_mut_slice` unsafe is to inform developers that they must take care of the updated_size after using it, and  at most of the time the develoers should use the safe `set_output` method.

crates/optee-utee/src/parameter/memref.rs
ivila: I think we can named it `get_buffer()`, and `as_mut_slice()` could be `get_buffer_mut()`

DemesneGH: Agree

examples/aes-rs/ta/src/main.rs
DemesneGH: `set_output` makes sense for me.
If there must use the `unsafe { output.as_mut_slice() }` and `update_size` pattern, how about adding some comment, and recommend users to use `set_output` as the first choice in their code?

DemesneGH:
LGTM, please feel free to merge once the CI passes, thanks!

PR #309, opened 2026-07-13
examples: reorganize crates into TA and CA workspaces

ivila:
This PR reorganizes the examples into dedicated CA and TA Cargo workspaces and consolidates shared protocol definitions into a single proto crate.

  ## Benefits

  ### Faster and more resource-efficient CI

  The workspace layout allows examples to reuse compiled dependencies instead of rebuilding them for each individual example.

  As a result:

  - Overall CI time decreased from 24 minutes to 10 minutes
  - CA build artifacts use 736 MB
  - TA build artifacts use 1.6 GB
  - Build artifacts are reused throughout CI and cleaned after the complete build

  ### More idiomatic Rust project structure

  The new layout follows standard Rust workspace conventions:

  - CA and TA crates are managed by separate workspaces
  - Protocol definitions are organized as modules in one shared proto crate
  - Common build logic is centralized in workspace-specific Makefile includes
  - Workspace-level dependency resolution and lock files reduce duplicated configuration

  This is especially useful for examples containing multiple TAs. Each TA can be represented as an independent workspace member while sharing dependencies, protocol types, build
  configuration, and artifacts naturally through Cargo.

  ### Better inter-TA communication

  The consolidated proto crate also improves inter-TA calls. A TA can directly import another TA’s UUID and command definitions from the corresponding protocol module.

  This removes the need to hard-code another TA’s UUID or duplicate its command IDs in the caller. Consequently:

  - UUIDs and commands have a single source of truth
  - Inter-TA callers and callees cannot silently drift apart
  - Adding or changing commands is safer and easier to maintain
  - Multi-TA dependencies are explicit in Rust code

  For example:

  use proto::hello_world::{Command, UUID};

  ## New Layout

  examples/
  ├── ca/
  │   ├── Cargo.toml
  │   ├── Makefile.include
  │   └── <CA crates>
  └── ta/
      ├── Cargo.toml
      ├── Makefile.include
      ├── proto/
      └── <TA crates>

  The resulting organization is faster to build, easier to maintain, safer for inter-TA communication, and better aligned with Rust conventions.


DemesneGH:
LGTM! BTW does this PR also help reduce building artifacts in OP-TEE repo? If so, could you help to pin the latest version in OP-TEE manifest after this PR merged? Thanks!

ivila:
> LGTM! BTW does this PR also help reduce building artifacts in OP-TEE repo? If so, could you help to pin the latest version in OP-TEE manifest after this PR merged? Thanks!

I’ll verify tomorrow and open a PR if it does (which I think it should).

PR #312, opened 2026-08-06
std: update Rust toolchain to nightly-2026-08-05

ivila:
- Pin rustc to nightly-2026-08-05 (1ed2df61) and libc to 0.2.189.
- Mark OP-TEE JSON targets as single-threaded and enable the new Cargo/rustc JSON target gates.
- Update cargo-optee and TA Makefiles so std builds pass json-target-spec and unstable-options without duplicating RUSTFLAGS.
- Fix nightly Clippy and macro expression compatibility in the SDK examples, including eth_wallet std builds.
- Fix FFI functions that return c_void.

### DIFF
--- a/aarch64-unknown-optee.json
+++ b/aarch64-unknown-optee.json
@@ -16,6 +16,6 @@
     "target-endian": "little",
     "target-pointer-width": 64,
     "vendor": "unknown",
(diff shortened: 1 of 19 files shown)


DemesneGH:
LGTM, thanks!

PR #321, opened 2026-09-16
optee-utee: make memref parameter access copy-based

ivila:
# Summary

Rework the typed memref parameter API so that TA code can no longer obtain a Rust slice directly over Normal-World shared memory. All reads and writes now go through explicit copies (`read_to_vec` / `read_at`, `set_output` / `write_at`), which removes the aliasing hazard and the easy TOCTOU window where a buffer is validated through one fetch and used through another.

# Motivation

A memref parameter is a `{buffer, size}` pair whose backing memory is mapped from the Normal World and stays writable by it for the whole call:
- Handing out `&[u8]` / `&mut [u8]` over that memory is unsound with respect to Rust aliasing (the REE mutates it behind the TA's back), and
- It invites double-fetch (TOCTOU) bugs: validate a field, then read it again "for real".
Copy-based access makes the safe thing the only thing the safe API exposes, and matches the guidance already in `docs/security-model.md` ("copy-then-validate").

# Changes
- `optee-utee: add ParameterMemref trait` — new supertrait exposing buffer_len(), shared by ParameterMemrefRead and ParameterMemrefWrite, replacing Read::buffer_len and Write::get_capacity. No behavior change.
- `optee-utee: remove get_buffer accessors` — drop the unsafe get_buffer / get_buffer_mut; add ParameterMemrefRead::read_at(offset, dest) returning the number of bytes read (BadParameters when offset is past the end, truncated at the end otherwise). read_to_vec remains the whole-buffer copy. Writes use set_output / write_at, which copy and update the reported size in one step.
- `examples, projects: use copy-based memref access` — migrate every TA (18 files) from get_buffer / get_buffer_mut / get_capacity to read_to_vec / read_at / set_output / buffer_len. Opportunistically adds missing input bounds checks in hotp-rs and signature_verification-rs.
- `docs: describe copy-based memref API` — update the parameter migration guide and the security model.
- `New unit test parameter::memref::tests::read_at_boundary_contract` covers the read_at boundary contract (offset == len → Ok(0), offset > len → BadParameters, no usize::MAX overflow).

# API / compatibility notes

- Breaking for the typed parameter API: **get_buffer, get_buffer_mut, and get_capacity are removed**, and buffer_len moves to the ParameterMemref supertrait. These types have not shipped in a release yet (they landed after v0.9.0), so no deprecation shim is provided. The legacy deprecated::ParamMemref::buffer() path is untouched.

# Future direction
Once **core::io / alloc::io** stabilize, I may re-express the `read/write` side as `io::Read` / `io::Write` adapters so callers can be generic over the standard traits. Any such adapter will keep the `copy` / `snapshot` semantics rather than exposing the shared buffer directly, since the REE can mutate it concurrently.


DemesneGH:
LGTM, thanks!

